PR #62, opened 2022-09-16
Square tiling pattern augmentation with example

nkemnitz:
Adds brightness augmentation to simulate stitching patterns, with example cue file


supersergiy (approved):
@nkemnitz Looks great, feel free to merge

PR #81, opened 2022-09-30
feat: RandomTranslateVolumetricIndex

nkemnitz:


supersergiy:
@nkemnitz I'll close this for now.
If you'd like to re-open this, it should be much easier to make it fit -- `IndexAdj` is not a thing now, a simple callable that takes and returns and Index is sufficient.

PR #100, opened 2022-11-16
Sequential + Module Primitives

dodamih:
New module primitives + sequential


supersergiy (changes requested):
@dodamih can you add the missing `#pragma: no cover`? good to merge after that

supersergiy (approved):
Nice work

PR #101, opened 2022-11-16
Basic trace suppress config

supersergiy:


dodamih:
Probably should change the filename to "online_finetuner" from "online_finetunner"

PR #103, opened 2022-11-16
Traceback logging improvement

supersergiy:


supersergiy:
Thanks @dodamih

supersergiy:
@dodamih You should check out the new traceback look after this is merged. Just import `zetta_utils` and make it throw an exception somewhere

PR #115, opened 2022-11-17
Merge convblock changes into main

dodamih:


supersergiy (changes requested):
You're just overwritting the file with your own. I fixed a couple of bugs in the main, so you should look into what you're ripping out

dodamih:
The two differences are the biases (https://github.com/ZettaAI/zetta_utils/blob/3e0e2278e5760968f56f6c94ade3cef94abc6a72/zetta_utils/convnet/architecture/convblock.py#L71) and removing defaultdict in favour of a regular dictionary with handling logic (https://github.com/ZettaAI/zetta_utils/blob/3e0e2278e5760968f56f6c94ade3cef94abc6a72/zetta_utils/convnet/architecture/convblock.py#L100). I'm not sure I see the reason for either change - can you explain the reason for these changes?

dodamih:
Ugh, let me do this properly

zetta_utils/convnet/architecture/convblock.py
```diff
@@ -44,9 +55,10 @@ def __init__(
         num_channels: List[int],
         activation: Callable[[], torch.nn.Module] = torch.nn.LeakyReLU,
         conv: Callable[..., torch.nn.modules.conv._ConvNd] = torch.nn.Conv2d,
-        padding_mode: Union[Literal["valid"], Literal["same"]] = "same",
         normalization: Optional[Callable[[int], torch.nn.Module]] = None,
         kernel_sizes: Union[List[int], int, List[Tuple[int, ...]], Tuple[int, ...]] = 3,
+        strides: Union[List[int], int, List[Tuple[int, ...]], Tuple[int, ...]] = 1,
+        paddings: Union[List[int], int, List[Tuple[int, ...]], Tuple[int, ...]] = 1,
```
supersergiy, line 61: What's the usecase for paddings instead of the padding mode? Do people actually use paddings other than `same` and `valid`?

dodamih: The intention was that in might be useful in cases where the convs are transposed convolutions or strided

supersergiy: Does `same` and `valid` do not what you want when using strided convolutions?

Fair point about transposed convs. I'm wondering if we should change the ConvBlock to only work for non-transposed convolutions, if it makes things simpler. Changing padding manually every time can be error prone.

dodamih: Strided conv don't work with the padding_mode I think:

"padding='valid' is the same as no padding. padding='same' pads the input so the output has the shape as the input. However, this mode doesn’t support any stride values other than 1."

supersergiy: I see. How about we do the same thing as torch does -- accept either these integers or `same`, `valid` strings? Then we can default to `same` while letting an adventurous user to make it work for all kinds of convs.
The parameter should definitely be called `padding` like you did, not `padding_mode`

dodamih: Sounds good

supersergiy (changes requested):
Actually, the typing for paddings is incorrect. It should be literals instead of str

dodamih:
So this type ignore comment thing has been showing up for me intermittently - each time this has happened, I just do git commit -n and then the next time I commit it doesn't throw an error. Any ideas?

dodamih:
Should be good now.

supersergiy (approved):
Thanks for fixing all the issues

PR #118, opened 2022-11-22
Parse neuroglancer layers

akhileshh:
* parse ngl layers
* create bounding boxes from annotations


zetta_utils/alignment/get_z_blocks.py
```diff
@@ -2,33 +2,20 @@
 
 from zetta_utils import builder, log
 from zetta_utils.bcube import BoundingCube
+from ..parsing.ngl_state import get_bcubes_from_annotations
+from ..parsing.ngl_state import load as load_ngl_layer
```
supersergiy, line 7: I think importing from `zetta_utils.parsing.ngl_state` would be slightly better here -- it will be slightly easier for a potential user to figure out where it's going or copy the import line to another file. 
Doing relative imports is definitely fine for within-module imports, but when we're importing from another module I think it's better to stick to absolute.

supersergiy (changes requested):
This looks good!
What do you think about the interface of `ngl_state` being two functions:
1. `read_remote_annotation` reads the layer, converts bounding boxes to bcubes, and points to Vec3D like things, and throws exception on any other object
2. `write_remote_annotation` takes a list of bcubes or Vec3D's and writes that as a remote annotation layer

All of the NG stuff would be nicely encapsulated, and it's general enough so that we would not need to touch this code again unless we need to support more annotation types.

akhileshh:
>  What do you think about the interface of ngl_state being two functions:

Yeah that sounds good. So the return type of `read_remote_annotations` would be `List[Union[BoundingCube, Vec3D]]`?

supersergiy:
I think `List[Union[BoundingCube, Vec3D]]` should be good -- maybe somebody will need a mix at some point.

supersergiy (approved):
This is really solid!

The only thing is that github shows that 108 files have changed, so it took some effort to figure out what actually changed. I think rebasing `main` instead of merging should make it better

akhileshh:
Actually this is the result of rebasing. There were multiple conflicts.

akhileshh:
closing in favor of #141

PR #132, opened 2022-12-02
pretty printing and alignment quality index

dodamih:


zetta_utils/mazepa_layer_processing/alignment/compute_alignment_quality.py
```diff
@@ -0,0 +1,174 @@
...
+
+logger = log.get_logger("zetta_utils")
+
+IndexT = TypeVar("IndexT", bound=LayerIndex)
+P = ParamSpec("P")
+R_co = TypeVar("R_co", covariant=True)
+
+
+@mazepa.taskable_operation
+def compute_misalignment_stats(**kwargs) -> Dict[str, Any]:
```
supersergiy, line 22: Can you get rid of the `**kwargs` here?

supersergiy (changes requested):
Looks good except for the `**kwargs`

zetta_utils/mazepa_layer_processing/alignment/compute_alignment_quality.py
dodamih: Yeah

dodamih: Wait, why is it better to get rid of kwargs? Is it dangerous? I think this is how your taskable operations were done before.

supersergiy: Arbitrary kwargs make code hard to read and hinder static typechecking.

I think the places you're referring to are probably wrapper functions that accept specific kwargs typed by ParamSpec.

supersergiy (approved):
💯

PR #143, opened 2022-12-05
high res encoding coarsener cue and regime

dodamih:
Cue updated to use new sequential rather than artificery


supersergiy (approved):
Looks good. Thanks for renaming the file so that many versions of coarsener can co-exist

PR #150, opened 2022-12-07
Dodam/vec3d as own type

dodamih:
zetta_utils/alignment/z_blocks.py still needs to be fixed since Union[SomeType, Vec3D] may not be supported, and there may be other things that are still broken that are not covered by the tests.


dodamih:
The failing test is not reproducible for me: `(zetta) dodam@Categoricity:~/zetta-packages/zetta_utils$ pytest tests/unit/layer/volumetric/cloudvol/test_backend.py 
============================================================================================= test session starts =============================================================================================
platform linux -- Python 3.10.4, pytest-7.1.1, pluggy-1.0.0
rootdir: /home/dodam/zetta-packages/zetta_utils, configfile: pyproject.toml
plugins: cov-3.0.0, mock-3.8.2, typeguard-2.13.3
collected 13 items                                                                                                                                                                                            

tests/unit/layer/volumetric/cloudvol/test_backend.py .............  `

zetta_utils/layer/volumetric/index.py
```diff
@@ -29,13 +29,13 @@ def default_convert(cls, idx_raw: RawVolumetricIndex) -> VolumetricIndex:  # pra
     def to_slices(self):
         return self.bcube.to_slices(self.resolution, self.allow_slice_rounding)
 
-    def pad(self, pad: Vec3D):
+    def pad(self, pad: Union[Vec3D, IntVec3D]):
```
supersergiy, line 32: How about making a Vec3D generic on the type of tuple element, and then defining a typealias `IntVec3D = Vec3D[int]`? Then `Union[Vec3D, IntVec3D] -> Vec3D`

zetta_utils/builder/parser.py
```diff
@@ -2,13 +2,17 @@
...
 
 from typeguard import typechecked
 
 from zetta_utils.partial import ComparablePartial
 
+from ..typing import IntVec3D, Vec3D
+
 REGISTRY: dict = {}
+REGISTRY_CAST_VEC3D: dict = {}
+REGISTRY_CAST_INTVEC3D: dict = {}
```
supersergiy, line 15: I'd prefer the `REGISTRY: dict[str, RegistryEntry]` approach, where
```python
@attrs.mutable
class RegistryEntry:
        fn: Callable
        name: str
        cast_vec3d: ...
```

zetta_utils/builder/parser.py
```diff
@@ -18,7 +22,12 @@
 
 
 @typechecked
-def register(name: str, versions=None) -> Callable[[T], T]:
+def register(
+    name: str,
+    cast_to_vec3d: Optional[List[str]] = None,
+    cast_to_intvec3d: Optional[List[str]] = None,
```
supersergiy, line 28: Is it a good idea to:
```python
default_converter={
   'resolution': Vec3D, 
   'bcube': BoundingCube.from_slices
}
``` 

How unlikely is it that we'd need to do this for new types? How worried are we about users abusing this?

tests/unit/layer/volumetric/cloudvol/test_backend.py
```diff
@@ -141,7 +141,7 @@ def test_cv_backend_write_scalar(clear_caches, mocker):
 
     index = VolumetricIndex(
         bcube=BoundingCube.from_slices((slice(0, 1), slice(1, 2), slice(2, 3))),
-        resolution=(1, 1, 1),
+        resolution=Vec3D((1, 1, 1)),
```
dodamih, line 144: This is pretty strange - I think I had the wrong version of the file checked out on my local branch and it never showed up on git status for some reason.

zetta_utils/layer/volumetric/index.py
dodamih: Considered it, but I thought stronger typing was better to prevent mistakes. For instance, there were a lot of places in the code where we were using Vec3D for things like chunk size and stride when we really meant IntVec3D. What you're suggesting is cleaner, though.

I'll see how well mypy does with the type checking when I do that.

zetta_utils/builder/parser.py
dodamih: Sounds good, though I guess how this is implemented depends on the answer to the next question.

zetta_utils/builder/parser.py
dodamih: I think it's unlikely that we'd need to do this for new types, but I think it is likely that we will want to do this everywhere once we start doing it to avoid `@builder` syntax. Not worried about users abusing it per se, but I'm worried that we will forget to write it for some functions or update it when we change the function signature and tracking down what's failing where is going to be difficult. If we are going to add a default converter, I'd prefer that it be done algorithmically from the type annotations somehow, though I don't know how feasible that is.

supersergiy: Ok, then how about we stick to a more generic syntax for now purely because it's fewer lines of code/fewer ifs/fewer tests:
```python
@builder.register(
    "foo", 
     default_converters={
           ....
     }
)
def foo(...)

supersergiy: Also, how about the following behavior changes:
1. `Vec3D((x, y, z)) -> Vec3D(x, y, z)
2. Auto conversion behaves differently based on the given value. If it's a dict, it's assumed to be `kwargs`. If it's a list, it's assumed to be `args`. Otherwise, it's assumed to be a complete object and auto conversion is not applied.

tests/unit/layer/volumetric/cloudvol/test_backend.py
supersergiy: I had the same. Then I noticed that the line numbers differ and re-cloned, and then it worked

zetta_utils/builder/parser.py
dodamih: Sounds reasonable. Let me see if I can make it work.

supersergiy (approved):
💯 🔥

supersergiy:
Looks like there's still some conflicts. Can you `rebase` this using git cli, without creating a merge commit?

PR #151, opened 2022-12-07
fix: make skips `Dict[str, int]` to allow deserialized dicts

supersergiy:
When dicts get deserialized from `cue/json/yaml`, the keys are always forced to be strings.
As the result `specs/examples/trainings/mimic_encoder.cue` wouldn't run on `main` as is.


dodamih (approved):
Oh, I broke this with the UNet commit. My bad.

supersergiy:
We both missed it...
I wonder what could've been done differently to avoid this. Breaking examples is rough

PR #154, opened 2022-12-08
Training auto resume

supersergiy:
Re-running the same spec will now automatically resume from the `last` checkpoint, if present.

Unfortunately it wasn't as easy as passing `ckpt_path="last"` to [trainer.fit](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.trainer.trainer.Trainer.html#pytorch_lightning.trainer.trainer.Trainer.fit) because that is actually not properly implemented for GCS checkpoints.


nkemnitz:
@supersergiy Is the `last` checkpoint always the end of an epoch, or is that name also used for checkpoints in the middle of an epoch? I saw a warning  when I explicitely tried to restart in the middle of an epoch: https://pytorch-lightning.readthedocs.io/en/1.6.0/advanced/fault_tolerant_training.html

Issue #136, opened 2022-12-03
NoSQL database integration


supersergiy:

supersergiy:
```python
l = build_kv_layer(backend=DataStoreBackend(), ....)

Read:
l[a: str] -> get the ‘value’ column for key “a”
l[(a: str, b: Tuple[str, ...])] -> get “b” columns for key “a” 
l[list[a: str], b: Tuple[str, ...]] -> batch read

write:
l[a: str] = value_BLOB
l[(a: str, b: Tuple[str, ...])] = {}
l[list[a: str], b: Tuple[str, ...]] = [{}]
```

supersergiy:
Closed via #157

PR #157, opened 2022-12-10
Implements generic db backend and index, with initial support for datastore

akhileshh:
#136


supersergiy (commented):
Looks pretty good. 
Could the backend logic be simplified if we treat the Value column exactly the same as all other columns?

akhileshh:
I suppose. Something like this? 
```
parent_key = Key("Row", row_key)
child_key = Key("Column", col_key, parent=parent_key)
entity = Entity(key=child_key)
entity[col_key] = <value>
```
where `col_key` can also be `"value"`.

supersergiy (approved):
Nice work! 💯

PR #188, opened 2022-12-18
feat: set up CPU/GPU cluster with node autoscaler support

nkemnitz:
* add a tiny separate node pool to run system pods for autoscaler
* add Taint `worker-pool` to CPU/GPU node pools to prevent system pods from getting scheduled on worker nodes
* add Toleration for `worker-pool` taint to mazepa generated deployments


supersergiy (approved):
💯

PR #189, opened 2022-12-19
chore: Update .dockerignore

nkemnitz:
Currently a copy of .gitignore. Only item removed is `.git`, which is currently not that large and could be useful for debugging an image.


supersergiy (approved):
💯

PR #202, opened 2023-01-02
Affine transform + 3D tile aug

supersergiy:


pyproject.toml
```diff
@@ -40,13 +40,14 @@ cli = [
 ]
 tensor-typing = [
     "torch == 1.12.1", # mypy error on newer
-    "numpy >= 1.23.1",
+    "numpy == 1.23.4",
```
nkemnitz, line 43: Why all the new pins?

nkemnitz (approved):

pyproject.toml
supersergiy: `boto3` related pins are necessary to build the docker with python3.9. Otherwise it can't resolve the environment. 
Hopefully we can unpin later after they figure out their stuff.
`numpy` pin is my mistake, I think. Will fix that. Thanks for checking!

PR #231, opened 2023-01-13
Dodam/subchunkable

dodamih:


zetta_utils/bcube/bcube.py
```diff
@@ -46,6 +48,22 @@ def ndim(self) -> int:
         """float of dimensions."""
         return len(self.bounds)
 
+    # NOTE: Cannot use VecT as return type since TypeVars are not callable.
+    @property
+    def start(self) -> Vec3D:  # pragma: no cover
```
supersergiy, line 53: Motion to kill the ND-ness of the BoundingCube

zetta_utils/bcube/bcube_strider.py
```diff
@@ -25,57 +27,209 @@ class BcubeStrider:
...
+    :param stride_start_offset: Where the stride should start in unit (not voxel),
+        modulo ``chunk_size``.
+    :param mode: The modes that can be chosen for the behaviour.
+        `shrink` is the default behaviour, which will round the bcube down to
+        be aligned with the ``stride_start_offset`` (or just the start of the bcube
+        if ``stride_start_offset`` is not set or ``stride`` != ``chunk_size``).
+        `expand` is similar to `shrink` except it expands the bcube.
+        `exact` will give full cubes aligned with the ``stride_start_offset`` and the
+        ``chunk_size``, as well as the partial cubes at the edges.
+    :param silent: Suppresses all warnings.
```
supersergiy, line 44: Is it better to let the default verbosity level handle this stuff? Maybe we can emit warnings as `info` or `debug`?

zetta_utils/layer/volumetric/backend.py
```diff
@@ -19,3 +22,35 @@ def enforce_chunk_aligned_writes(self) -> bool:
...
+
+    @abstractmethod
+    def get_voxel_offset(self, resolution: Vec3D) -> IntVec3D:
+        ...
+
+    @abstractmethod
+    def set_voxel_offset(self, voxel_offset: IntVec3D, resolution: Vec3D) -> None:
+        ...
+
+    """ Sets the voxel offset at the given resolution for the backend.
```
supersergiy, line 34: This comment should go instead of `...`?

zetta_utils/layer/volumetric/cloudvol/backend.py
```diff
@@ -213,3 +235,71 @@ def clone(self, **kwargs) -> CVBackend:
...
+
+    def assert_idx_is_chunk_aligned(self, idx: VolumetricIndex) -> None:
+
+        # check that the idx given is chunk_aligned, and give suggestions
+        idx_expanded = self.get_chunk_aligned_index(idx, mode="expand")
+        idx_rounded = self.get_chunk_aligned_index(idx, mode="round")
+        idx_shrunk = self.get_chunk_aligned_index(idx, mode="shrink")
+
+        if idx != idx_expanded:
+            raise ValueError(
```
supersergiy, line 295: Could we add a test for this? This error condition sounds actually important.

zetta_utils/layer/volumetric/index.py
```diff
@@ -19,44 +19,85 @@ class VolumetricIndex:  # pragma: no cover # pure delegation, no logic
...
+
+    @property
+    def shape(self) -> IntVec3D:
+        return self.stop - self.start
+
     def to_slices(self):
         return self.bcube.to_slices(self.resolution, self.allow_slice_rounding)
 
-    def pad(self, pad: IntVec3D):
+    def pad(self, pad: IntVec3D) -> VolumetricIndex:
```
supersergiy, line 37: should we rename these guys to `with_pad`, `with_translation`? or maybe `padded`, `translated`?

zetta_utils/mazepa_layer_processing/common/blendable_apply_flow.py
```diff
@@ -69,51 +295,105 @@ def make_tasks_without_blending(
...
-
-        tasks = []
-        idx_chunks_temps: List[List[VolumetricLayer]] = [[] for _ in idx_chunks]
-        for chunker, chunker_idx in self.chunker.split_into_nonoverlapping_chunkers(
+    ) -> Tuple[
+        List[mazepa.tasks.Task[R_co]],
+        List[mazepa.tasks.Task[None]],
+        List[List[VolumetricIndex]],
+        List[List[VolumetricLayer]],
+        List[List[VolumetricLayer]],
```
supersergiy, line 309: This feels like a lot of of lists, hard to understand which is which.
Would it be better to declare a small class up above, with 5 named fields, and return an instance of that class?

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -0,0 +1,204 @@
...
+        blend_pads,
+        blend_modes,
+        temp_layers_dirs,
+    ]:
+        if isinstance(arg, list):
+            if not len(arg) == num_levels:
+                raise ValueError(
+                    "The arguments `processing_chunk_sizes`, `max_reduction_chunk_sizes`"
+                    " (optional), `blend_pads`, `crop_pads`, `blend_modes`, `temp_layers_dirs`"
+                    " must be singletons or lists where the longest lengths match."
```
supersergiy, line 111: Not sure what `lists where the longest lengths match` means here. Isn't this checking that all the lengths are equal to `processing_chunk_sizes`? Feels odd to include `processing_chunk_sizes` itself in the check.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -0,0 +1,204 @@
...
+
+IndexT = TypeVar("IndexT")
+P = ParamSpec("P")
+R_co = TypeVar("R_co", covariant=True)
+T = TypeVar("T")
+
+
+@mazepa.taskable_operation_cls
+@attrs.mutable
+class DelegateBlendOperation(Generic[P, R_co]):
```
supersergiy, line 27: How about `DelegatedSubchunkedOperation`?

zetta_utils/training/datasets/joint_dataset.py
```diff
@@ -39,13 +39,13 @@ def __attrs_post_init__(self):
             for key in self.datasets.keys():
                 if num_samples == len(self.datasets[key]):
                     logger.warning(
-                        f"JointDataset: Dataset {key} has {len(self.datasets[key])} samples, "
+                        f"JointDataset: Dataset '{key}' has {len(self.datasets[key])} samples, "
```
supersergiy, line 42: Thanks for doing this

supersergiy (approved):
Awesome! Left some small comments, up to your digression

zetta_utils/bcube/bcube.py
```diff
@@ -83,7 +101,7 @@ def from_coords(
         cls,
         start_coord: VecT,
         end_coord: VecT,
-        resolution: VecT,
+        resolution: Optional[VecT] = None,
```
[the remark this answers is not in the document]
dodamih: It almost does the same thing as setting the default to be (1, 1, 1), which is probably cleaner. The reason why it's set to `None` is because VecT is not known - it can be a Sequence or a Vec3D. I can rebase to incorporate Nico's change and then set the default to `Vec3D(1,1,1)`

zetta_utils/bcube/bcube.py
dodamih: As noted in the comment, I have to be able to construct the Vec3D to return it, and VecT isn't callable. This is probably ugliness that we have to accept until the BoundingCube is changed.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
dodamih: Yeah, that sounds good.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
dodamih: Oops, sorry - forgot to change that when I changed the code.

zetta_utils/layer/volumetric/cloudvol/backend.py
dodamih: There is already testing for this - not sure why codecov is barfing. See line 252: https://github.com/ZettaAI/zetta_utils/pull/231/files#diff-bc8abc2012f5ca7148a4aea069752161f185943809f9b30faf828d45611dbc28

zetta_utils/layer/volumetric/backend.py
dodamih: Oops.

zetta_utils/bcube/bcube_strider.py
dodamih: I'm not sure - part of me thinks that any time the index changes inside a function we should emit an explicit warning, but using info or debug might be better. Your call.

zetta_utils/layer/volumetric/index.py
dodamih: Yeah, I can do it.

zetta_utils/training/datasets/joint_dataset.py
dodamih: My pleasure

zetta_utils/mazepa_layer_processing/common/blendable_apply_flow.py
dodamih: I don't think the overhead (both for the person having to dig through to find the class and from a performance standpoint) is worth it given that it's only used twice; I think a class might be necessary when arbitrary blending is implemented, but for now I'd say no.

zetta_utils/bcube/bcube.py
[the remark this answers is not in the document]
dodamih: Oh, even with Nico's changes this doesn't work because VecT is not known at runtime. Not sure what's a good way to do this. Probably another "apply alchemy to BoundingCube and the problem will go away" item.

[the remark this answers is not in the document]
supersergiy: How about we keep Bcube as is to merge this in ASAP, but then put on a high priority issue to make a 3D-only bbox?

zetta_utils/layer/volumetric/index.py
dodamih: Actually, on second thought, I think this is fine as is. The threat if we switch to `with_pad` is that the user might think that the pad is an option that's set for whenever some operation is done (like a preprocessor).  Thoughts?

zetta_utils/bcube/bcube_strider.py
supersergiy: I think it's better to not have ad-hoc ways of handling verbosity

zetta_utils/layer/volumetric/cloudvol/backend.py
supersergiy: The link doesn't take me to the file. Can we confirm that that exact line is getting hit by tests by putting some `breakpoint()` or assert in there? codecov doesn't fail often

zetta_utils/layer/volumetric/index.py
supersergiy: Maybe `padded` then?
Are we going to switch to `with_index_processors` for immutable layers? If so, the `with_` thing can be consistent, and users will know what to expect.

zetta_utils/bcube/bcube.py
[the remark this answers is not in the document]
dodamih: Sounds good.

zetta_utils/bcube/bcube_strider.py
dodamih: Yeah, I guess if it has weird behaviour the user can just run `-v`. I'll change them to info.

zetta_utils/layer/volumetric/index.py
dodamih: I think `padded` / `translated` is clearest for this usage (even outweighing the consistency). I'll change to those for both the bcube and the index.

zetta_utils/layer/volumetric/cloudvol/backend.py
dodamih: I think it might have been catching a different `ValueError` - testing now.

dodamih: Yeah, that was it.

PR #249, opened 2023-01-23
feat: caching, lambda functions, arbitrary blending, info caching

dodamih:


zetta_utils/mazepa_layer_processing/common/volumetric_callable_operation.py
```diff
@@ -51,30 +51,28 @@ def with_added_crop_pad(self, crop_pad: IntVec3D) -> VolumetricCallableOperation
         return attrs.evolve(self, crop_pad=self.crop_pad + crop_pad)
 
     def __attrs_post_init__(self):
-        output_crop_raw = self.crop_pad / self.res_change_mult
-        for e in output_crop_raw:
+        input_crop_pad_raw = self.crop_pad / self.res_change_mult
```
supersergiy, line 54: Shouldn't this be `input_crop_pad_raw = self.crop_pad * self.res_change_mult`?

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -50,144 +52,164 @@ def expand_if_singleton(arg: Union[T, List[T]], length: int) -> List[T]:
 
 
 @builder.register("build_subchunkable_apply_flow")
-def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg, line-too-long, too-many-locals, too-many-branches
-    bbox: BBox3D,
+def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg, line-too-long, too-many-locals, too-many-branches, too-many-statements
+    start_coord: IntVec3D,
+    end_coord: IntVec3D,
+    coord_resolution: Vec3D,
```
supersergiy, line 58: If subchunkable is created directly by the user, then this is great. 
But if it's created from within another flow ([[eg](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/mazepa_layer_processing/alignment/compute_field_flow.py#L158)]), converting `idx.bbox` to `start_coord`, `end_coord` and `coord_resolution` doesn't sound clean. 

Perhaps `build_subchunkable_apply_flow_shortcut` can be the one registered with the builder and take in coords, build coords into a bbox and pass it into `build_subchunkable_apply_flow`?

If this has non-trivial concerns/consequences, we can defer it to a separate PR after this is merged.

supersergiy (changes requested):
I think there might be a bug in the input->output crop conversion, but really solid otherwise

zetta_utils/mazepa/execute.py
```diff
@@ -126,6 +130,7 @@ def _execute_from_state(
         expected_operation_counts = {}
 
     with ExitStack() as stack:
+        #        show_progress = False
```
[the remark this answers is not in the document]
dodamih: Oops, yeah, sorry about that

zetta_utils/mazepa_layer_processing/common/volumetric_callable_operation.py
dodamih: You're correct, will fix.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
dodamih: I think having "shortcut" sounds even messier - I'll change it so that it can gracefully accept either `bbox` or `start_coord, end_coord, coord_resolution`.

pyproject.toml
```diff
@@ -93,6 +93,8 @@ training = [
     "torch == 1.12.1", # mypy error on newer
     "pytorch-lightning == 1.7.7", # mypy error on newer
     "wandb >= 0.13.1",
+    "onnx >= 1.13.0",
+    "onnxruntime-gpu >= 1.13.1",
```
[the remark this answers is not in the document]
dodamih: Yeah, I'll add it and move it

PR #266, opened 2023-01-29
fix: upgrade isort

akhileshh:


akhileshh:
something similar for reference - [SO post](https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co)

supersergiy:
Thanks for the fix!

PR #267, opened 2023-01-29
Resource management - tracking and garbage collection

akhileshh:


zetta_utils/mazepa_addons/execution_tracker.py
```diff
@@ -0,0 +1,24 @@
+import os
+import time
+
+from zetta_utils.layer.db_layer import RowDataT, build_db_layer
+from zetta_utils.layer.db_layer.datastore import DatastoreBackend
+
+DEFAULT_PROJECT = "zetta-research"
+EXECUTION_DB_NAME = "execution-info"
+EXECUTION_DB = build_db_layer(
```
supersergiy, line 9: I think you're missing `__init__` file which imports this

zetta_utils/mazepa_addons/resource_allocation/tracker.py
```diff
@@ -0,0 +1,32 @@
...
+
+@attrs.frozen
+class ExecutionResource:
+    execution_id: str
+    resource_type: str
+    resource_name: str
+
+
+def register_execution_resource(resource: ExecutionResource) -> None:
+    _resource = attrs.asdict(resource)
```
supersergiy, line 29: You were able to bypass the issue without the type ignore?

supersergiy (approved):
Great work 💯 
`__init__.py` will fix the coverate

supersergiy:
Also, can you rename one of `tracker.py` and `execution_tracker.py`? 
It's a little ambiguous what is where

zetta_utils/mazepa_addons/resource_allocation/tracker.py
akhileshh: I realized the previous code was incorrect because the user index was missing col_keys. But yeah somehow mypy doesn't complain here now.

However it does complain in `update_execution_info` if I remove the `RowDataT` type hint.

akhileshh:
> Also, can you rename one of `tracker.py` and `execution_tracker.py`?

My reasoning was `tracker.py` is part of  `resource_allocation` so its for resource tracking, the other for execution tracking.
But will make them explicit.

supersergiy (dismissed):
Very nice solution.
For now GC will assume single project, single cluster?

supersergiy (approved):

PR #276, opened 2023-02-01
fix: restore 7f8abe2

nkemnitz:
The Sequence type condition got lost somewhere in #245.
mypy doesn't complain about the missing `T_co`, but pylance does in a cryptic `Cannot create consistent method ordering` message.


supersergiy:
Huh that's interesting. I used to get runtime errors on python3.10 when`[T_co]`, but I guess CI says there's no problem

PR #278, opened 2023-02-10
feat: ChainIndexer to chain SampleIndexers

nkemnitz:
A similar behavior would also be possible with Vertical Joint Datasets, but having a dedicated chain indexer should help keep the spec files more readable.


supersergiy (approved):
This looks pretty cool!
Since other dataset utilities are documented, do you mind including a pointer to this in the [[docs](https://github.com/ZettaAI/zetta_utils/blob/akhilesh/resource-management/docs/source/zetta_utils.training.rst)] before merging?

PR #286, opened 2023-02-14
New specs and updated BaseEncode regime

nkemnitz:
First commit contains the new specs used for training/inference for what we have right now, so that I don't lose them. Also adds the coarsener regime (training+inference)

Second, separate commit updates the BaseEncoder regime - no fundamental change, just slightly different handling of boundary regions. But wanted to check with you first If that is problematic,


supersergiy:
@nkemnitz is this still relevant?

supersergiy (approved):
This looks good! The conflicts should be related to the little refactoring I did -- I moved alignment related regimes to `regimes/alignment`.

PR #293, opened 2023-02-16
fix: make `ffspec`-based info fetching fork safe

supersergiy:


nkemnitz (approved):
Seems to work pretty reliable with 4 workers. Not as reliable with 20, but we determined that this was already the case before b90a6916b96c60d87c7833004e963d0949035a92

Other workaround: Set only validation worker_num to 0

PR #296, opened 2023-02-20
feat: perlin noise generator

nkemnitz:


nkemnitz:
The generator spec is only semi-working: I think I am using the flows incorrectly - it's only calling the noise generator function once, and uploading the same result to all 10 sections.

specs/examples/inference/gen_perlin_noise.cue
```diff
@@ -0,0 +1,73 @@
...
+					device: "cuda"
+				},
+				{
+					"@type": "rand_perlin_2d_octaves"
+					input: [1024, 1024]
+					res: [8, 8]
+					octaves: 6
+					device: "cuda"
+				}
+			]
```
supersergiy, line 36: This is saying -- "for each given bounding cube, rearrange the following data". This phrasing makes it so that the data is generated only once, and then same data is rearranged over and over. 

One way to fix this is to have `rand_perlin_2d_octaves` produce the channel dimension. What do you think?

supersergiy (changes requested):

specs/examples/inference/gen_perlin_noise.cue
nkemnitz: Adding support for multiple channels is a good suggestion, might even boost performance. It doesn't eliminate the need for rearrange, though? Output is `CXY` (or `NCXY`) and CloudVolume expects `CXYZ`.
I guess I need a dedicated module to do the reshaping and function calling, no way to just do it within the cue spec?

supersergiy: If perlin noise generates `NCXY`, a working solution would be to apply rearrange as a write post processor.

At the same time, `CXYZ` is a general `zetta_utils` convention. Whenever we have tensor ops that work with more channels than just XY, we always assume `CXYZ` in and out (eg [[masks.py](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/tensor_ops/mask.py)]). To keep up the convention, I think it would be better to have the perlin noise generator return `CXYZ(N)` than `NCXY`.

Alternatively, we could add a wrapper function that takes in a _callable_ that produces XY data and calls it several times to fill the `C` and potentially `Z` dimensions.

supersergiy (approved):
💯

PR #297, opened 2023-02-20
Feat: UNet extensions 

nkemnitz:
* minor fix regarding downsample/upsample calls
* UpConv layer
* switch between residual and concatenated skip connections

* breaking: skip connection should occur before first upsample layer


supersergiy (commented):
The code looks good to me. With both `.jit` stuff being broken and no versioning, I don't think we can merge this just yet.

PR #298, opened 2023-02-21
feat: stable random indexer

supersergiy:


nkemnitz (approved):
Could you please update the description of the function to reflect that it now _does_ cover all samples and guarantees uniqueness? ``PieceIndexer`` is now also a ``SampleIndexer``.

Otherwise good to merge.

PR #318, opened 2023-02-27
fix: unsqueeze_input_to defaults to None

nkemnitz:


nkemnitz:
Current behavior doesn't match the spec, but also breaks compatibility with existing specs.

supersergiy:
This used to be None, but in that case it doesn't fit `zetta_utils` ZCXY convention.
Should we update documentation instead?

PR #336, opened 2023-03-23
feat: TensorStore for local layers

dodamih:


zetta_utils/layer/volumetric/cloudvol/backend.py
```diff
@@ -299,6 +306,12 @@ def write(self, idx: VolumetricIndex, data: torch.Tensor):
         cvol[slices] = data_final
         cvol.autocrop = False
 
+    def as_type(self, backend_type) -> VolumetricBackend:  # pragma: no cover # type: ignore
+        type_args = [f.name for f in attrs.fields(backend_type)]
+        keys_to_use = set(dir(self)).intersection(type_args)
+        kwargs_to_use = {k: getattr(self, k) for k in keys_to_use}
+        return backend_type(**kwargs_to_use)
```
supersergiy, line 313: Why don't we just have a function `to_tensorstore_backend` that takes a volumetric backend and returns `tensorstore` backend? Using inheritance here leads to ugly typing.

Also, does the following not work here?
```python
from typing import Type

T = TypeVar('T', bound=VolumetricBackend)

def as_type(self, backend_type: Type[T]) -> T:
   ...
```

zetta_utils/layer/volumetric/tensorstore/backend.py
```diff
@@ -0,0 +1,277 @@
...
+        the layer info already exists.
+
+    """
+
+    path: str
+    info_spec: Optional[PrecomputedInfoSpec] = None
+    on_info_exists: InfoExistsModes = "expect_same"
+
+    def __attrs_post_init__(self):
+        if self.info_spec is not None:
```
supersergiy, line 63: Can this be somehow factored to avoid code duplication with `CVBackend`?

zetta_utils/layer/volumetric/tensorstore/backend.py
```diff
@@ -0,0 +1,277 @@
...
+                    raise RuntimeError(
+                        f"Info created by the info_spec {self.info_spec} is not equal to "
+                        f"info existing at '{self.path}' "
+                        "while `on_info_exists` is set to 'expect_same'"
+                    )
+                if existing_info != new_info:
+                    self.info_spec.write_info(self.path)
+            self.info_spec.reference_path = self.path
+            _info_cache[_info_hash_key(self.path)] = new_info
+            _ts_cache.clear()
```
supersergiy, line 85: Wouldn't it be sufficient to just delete `self.path` from `_ts_cache`?

supersergiy (changes requested):

zetta_utils/layer/volumetric/tensorstore/backend.py
dodamih: Yeah - cachetools didn't support a clean way to do it which is why I'm just deleting the whole thing, but for bunk >> punk situation that's necessary. I'll implement it correctly before I push again.

zetta_utils/layer/volumetric/cloudvol/backend.py
dodamih: The reason was that I wanted to avoid a circular import and didn't want to hardcode `if check_type(layer.backend, CVBackend) or check_type(layer.backend, VolumetricLayerSetBackend): layer.backend.to_tensorstore_backend` in volumetric_apply_flow. I'll factor out the PrecomputedInfoSpec (alongside the `__attrs_post_init__` below), and just add it as a hardcode (rather than a file-level variable).

The typing should work if I'm not inheriting.

zetta_utils/layer/volumetric/tensorstore/backend.py
supersergiy: It's representing exactly the same logic though. How likely is it that one of them will change in a way separate from the other?

zetta_utils/layer/volumetric/cloudvol/backend.py
dodamih: Hmm, I think that the typing cannot actually be fixed because of the `VolumetricSetBackend` unless we want to make a VolumetricSetBackend be a subtype of TSBackend. I'll leave `to_tensorstore_backend` as something that returns `VolumetricBackend` even though that's not perfect.

PR #341, opened 2023-03-28
feat: TransientError handling for mazepa executions

supersergiy:
Lets the user specify which Exceptions/Text indicate a transient error and should be retried, rather than aborting execution.

These specification are passed to `@taskable_operation`/`@taskable_operation_cls` decorators to allow fine grained control over which tasks should be retried in which conditions. Some reasonable default set of transient exceptions is provided.

Transient errors will be ignored only if the task has been attempted less than a certain number of times, and will error out aborting the execution after that. This is because some errors (aka NVIDIA driver) can be both transient or non-transient.

### DIFF
--- a/zetta_utils/mazepa/tasks.py
+++ b/zetta_utils/mazepa/tasks.py
@@ -9,9 +9,11 @@ from concurrent.futures import TimeoutError as PebbleTimeoutError
 from typing import (
     Any,
     Callable,
+    Final,
     Generic,
     Iterable,
     Protocol,
+    Sequence,
     Type,
     TypeVar,
     cast,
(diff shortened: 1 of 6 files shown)


zetta_utils/mazepa/tasks.py
```diff
@@ -96,13 +119,22 @@ def __call__(self, debug: bool = True) -> TaskOutcome[R_co]:
             run_completion_callbacks = True
             self.status = TaskStatus.SUCCEEDED
         else:
-            run_completion_callbacks = True
-            self.status = TaskStatus.FAILED
-            if isinstance(self.outcome.exception, PebbleTimeoutError):
-                logger.debug(f"Task {self.id_} execution timed out!")
-                # Abstract the pebble library from the user
+            if self.curr_retry < self.max_transient_retry and max(
```
nkemnitz, line 122: `any` is a better fit than `max`

nkemnitz (approved):
This NVIDIA driver exception is weird... Pretty sure it happens during preemption, but I don't understand why.
GCE will send a preemption notice (30s) to GKE, which is caught by kubelet, and in turn is supposed to give every non-critical pod 15s graceful shutdown time [[Link](https://cloud.google.com/kubernetes-engine/docs/how-to/preemptible-vms#graceful-shutdown)]
But even if the pod does not finish the task within 15s, it should just forcefully shut down the pod (resulting in a timeout error for that task, I guess), instead of waiting until the GPU becomes unavailable causing some cryptic error message?...

PR #348, opened 2023-03-30
feat: subchunkable progress shows proper name, operation summary

dodamih:


supersergiy (approved):
I think it would be slightly cleaner to find out the OP name within the delegation task by looking into what it wraps recursively, but this is fine

PR #356, opened 2023-04-06
fix(mazepa): Another Runtime error caused by expired grace period

nkemnitz:
Still considering this a temporary fix. Both exceptions are legitimate and should not be retried when they occur on an instance without GPU or incorrect CUDA setup.

### DIFF
--- a/zetta_utils/mazepa/tasks.py
+++ b/zetta_utils/mazepa/tasks.py
@@ -56,7 +56,14 @@ class TransientErrorCondition:
 
 DEFAULT_TRANSIENT_ERROR_CONDITIONS: Final = (
     TransientErrorCondition(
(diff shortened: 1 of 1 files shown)


supersergiy:
This looks good other than the line too long

nkemnitz:
Btw, was it intentional to make those conditions members of every individal task? It's bloating the messages a bit.

supersergiy:
It's good to have task level control of the behavior. A certain exception could be transient for one task type but mean failure for another task type. I'm couldn't rule out the possibility that something considered "default transient" will be something we don't want to ignore for some task...

If we're sure that defaults always apply, we can make the tasks carry only additional exceptions.

supersergiy (approved):

PR #358, opened 2023-04-07
optimisation: remove quadratic scheduling

dodamih:
Came up with a much cleaner way to do it after I wrote a lot of code for splitting into 3D; I'm leveraging the fact that the chunks are in row-major order already.

### DIFF
--- a/zetta_utils/layer/volumetric/tools.py
+++ b/zetta_utils/layer/volumetric/tools.py
@@ -117,34 +117,11 @@ class VolumetricIndexChunker(IndexChunker[VolumetricIndex]):
         idx: VolumetricIndex,
         stride_start_offset_in_unit: Optional[Vec3D[int]] = None,
         mode: Literal["shrink", "expand", "exact"] = "expand",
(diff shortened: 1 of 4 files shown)


zetta_utils/layer/volumetric/tools.py
```diff
@@ -117,23 +117,54 @@ def __call__(
...
+    ) -> Vec3D[int]:  # pragma: no cover
+
+        return self._get_bbox_strider(idx, stride_start_offset_in_unit, mode).shape
+
+    def _get_bbox_strider(
+        self,
+        idx: VolumetricIndex,
+        stride_start_offset_in_unit: Optional[Vec3D[int]] = None,
+        mode: Literal["shrink", "expand", "exact"] = "expand",
+    ) -> BBoxStrider:  # pragma: no cover
```
supersergiy, line 149: This function has 3 if-else clauses, so `#pragma: no cover` shouldn't be used here. It's an internal function, so we should _not_ have a test for it, but tests for its clients (aka `self.__call__`) should cover all the branches here.

supersergiy (changes requested):

PR #362, opened 2023-04-10
hotfix: handle cases where shape is 1 in reduction chunks

dodamih:


zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -172,6 +173,7 @@ def get_blending_weights(  # pylint:disable=too-many-branches, too-many-locals
         *idx_subchunk.aligned(idx_roi),
     )
 
+    print(idx_subchunk.aligned(idx_roi))
```
supersergiy, line 176: print has to go

supersergiy (changes requested):

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
dodamih: Yeah, just saw that

PR #380, opened 2023-04-25
feat: unflatten subchunkable

dodamih:


zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -605,8 +606,8 @@ def build_volumetric_apply_flow(  # pylint: disable=keyword-arg-before-vararg
     intermediaries_dir: Optional[str] = None,
     allow_cache: bool = False,
     clear_cache_on_return: bool = False,
-    *args: P.args,
-    **kwargs: P.kwargs,
+    op_args: Optional[tuple] = None,
```
supersergiy, line 609: Is there a problem with this being an Iterable instead of Tuple?

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -624,6 +625,10 @@ def build_volumetric_apply_flow(  # pylint: disable=keyword-arg-before-vararg
         allow_cache=allow_cache,
         clear_cache_on_return=clear_cache_on_return,
     )
-    flow = flow_schema(idx, *args, **kwargs)
+    if op_args is None:
```
supersergiy, line 628: We can just assign the default value to be (), no need to go through None here because tulles are immutable.

We could also set kwargs to be a mapping and set the default to some implementation of an immutable mapping. Something like https://github.com/MagicStack/immutables

The downside is that the default here will depend on some third part library stuff, which could make it confusing to the reader. What do you think?

supersergiy (changes requested):

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
dodamih: No - I can change it.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
dodamih: I think simpler is better - it'll probably be confusing for future readers, and there isn't that much that we get out of it given that kwargs are dictionaries in 99.9% of situations. Do you foresee a use case where we'll use some arbitrary mapping kwargs instead?

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
dodamih: Oh, wait - there is an issue indeed. If it's Iterable, it doesn't support `len`.

supersergiy: If `len` is necessary, we can try [[Collection](https://docs.python.org/3/library/collections.abc.html)]

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
supersergiy: Using `Mapping` instead of `dict` is just more general, we can't anticipate all of the users crazy ideas.

What I meant with the `immutables` is to change
```python
op_kwargs: Optional[Mapping[str, Any]] = None,

...
if op_kwargs is None:
   op_kwargs = {}
```
into 
```python
op_kwargs: Mapping[str, Any]] = immutables.Map(),
...
```

This way we can avoid the extra if statement here and in other analogous places.

dodamih: Python does have a built-in Mapping that's immutable, so let me try using that.

supersergiy (approved):
💯

PR #381, opened 2023-04-27
feat: Vec3D is hashable

dodamih:


supersergiy:
Isn't it dangerous to have it both hashable and mutable?

dodamih:
It is dangerous, but I think it's okay for our use cases. I could add a warning as I suggested today if you think it's necessary (first time hash is called, keep track of what the value was at that point).

supersergiy:
I think somebody can get really screwed over if we do this. Why don't we just deprecate `__setitem__` to stay on the safe side?

dodamih:
`__setitem__` isn't the issue as much as `attrs.frozen` not working with `init=False`, meaning that using numpy numbers directly might not work.

supersergiy:
But at the very least we should disable `__setitem__`, right?

supersergiy:
What does `attrs` not like with `frozen(init=False)`?

dodamih:
You cannot do things like `self.x = var` when you have `attrs.frozen` because it overwrites the `setattr`: https://github.com/python-attrs/attrs/issues/113

supersergiy:
Does `object.setattr` trick not work? https://github.com/ZettaAI/zetta_utils/blob/bb10f113b8858754253458d801902b4ccbeac34d/zetta_utils/geometry/bbox_strider.py#L59

dodamih:
Oh, that trick did work. Strange.

dodamih:
Should be ready now.

supersergiy (changes requested):
fails CI

supersergiy (approved):

supersergiy (approved):

PR #390, opened 2023-05-05
feat: integration tests with Git LFS

dodamih:
This commit does not fix the multilevel writeprocs yet, but the tests for it are there.

The output comparison will tell you which files differ - I figured that there is little sense in returning the difference as tensors since the debugger will likely need to look at the volumes in neuroglancer anyway.


supersergiy (approved):
730 files! Has to be a record for us.

Looks good. How can we trigger integration tests through GitHub Actions?

PR #391, opened 2023-05-08
feat(training): adds ddp multinode support

akhileshh:
* refactors infrastructure code to `cloud_management` module
* adds ability to launch ddp training as a k8s job; without sqs


zetta_utils/cloud/resource_allocation/k8s/job.py
```diff
@@ -0,0 +1,143 @@
...
+
+        try:
+            yield
+            while True:
+                job = batch_v1_api.read_namespaced_job_status(
+                    name=job.metadata.name,
+                    namespace=namespace,
+                )
+                logger.info(f"`{job.metadata.name}` job status: {job.status.ready}")
+                if job.status.ready == 1:
```
supersergiy, line 115: Is ready a boolean? If it is, this should be `if job.status.ready:`

zetta_utils/cloud/resource_allocation/k8s/job.py
```diff
@@ -0,0 +1,143 @@
...
+                namespace=namespace, label_selector=f"job-name={job.metadata.name}"
+            )
+            job_pod = pods.items[0]
+            log_stream = watch.Watch().stream(
+                core_api.read_namespaced_pod_log,
+                name=job_pod.metadata.name,
+                namespace=namespace,
+            )
+            for output in log_stream:
+                logger.info(output)
```
supersergiy, line 130: I don't really understand this context manager. It yields control to the user, and then starts monitoring the status of the job in the teardown code, which makes it seem like this should not be a context manager at all, but rather a just a plain blocking function.

zetta_utils/cloud/__init__.py
supersergiy, line 1: `cloud` feels like too broad of a name that may be confusing to users. There are many other parts of the codebase that deal with cloud and do not live in this folder.

Should we rename it to `cloud_resource_management`?

zetta_utils/training/lightning/train.py
```diff
@@ -63,3 +83,69 @@ def lightning_train(
         val_dataloaders=val_dataloader,
         ckpt_path=ckpt_path,
     )
+
+
+@builder.register("lightning_train_remote")
+@typeguard.typechecked
+def lightning_train_remote(image: str, resources: dict, spec_path: str) -> None:
```
supersergiy, line 90: User should be able to choose the cluster rather than falling back on default. We often juggle clusters and projects, and it's important to be able to do that without code chages.

zetta_utils/training/lightning/train.py.bkp
supersergiy, line 1: This should be cut, right?

zetta_utils/training/lightning/train.py
```diff
@@ -2,21 +2,41 @@
...
+
+DEFAULT_GCP_CLUSTER_NAME: Final = "zutils-x3"
+DEFAULT_GCP_CLUSTER_REGION: Final = "us-east1"
+DEFAULT_GCP_CLUSTER_PROJECT: Final = "zetta-research"
+
+DEFAULT_GCP_CLUSTER: Final = resource_allocation.k8s.ClusterInfo(
+    name=DEFAULT_GCP_CLUSTER_NAME,
+    region=DEFAULT_GCP_CLUSTER_REGION,
+    project=DEFAULT_GCP_CLUSTER_PROJECT,
+)
```
supersergiy, line 26: These constants are repeated twice, once here and once in `execute_on_gcp_with_sqs`. They should probably move somewhere like `cloud.constants.py`

supersergiy (changes requested):

zetta_utils/cloud/resource_allocation/k8s/job.py
akhileshh: no it's `int`

zetta_utils/cloud/resource_allocation/k8s/job.py
akhileshh: Tried but it prevented the `configmap_ctx` from being `enter`'d, the `pod` was forever stuck in `ContainerCreating` state waiting for the `configmap`

zetta_utils/cloud/resource_allocation/k8s/pod.py
```diff
@@ -0,0 +1,87 @@
...
+
+from kubernetes import client as k8s_client  # type: ignore
+from zetta_utils import log
+
+from .secret import get_worker_env_vars
+
+logger = log.get_logger("zetta_utils")
+
+
+def get_pod_spec(
```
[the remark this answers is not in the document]
akhileshh: `spec` is also a `k8s` term though

zetta_utils/cloud/__init__.py
akhileshh: Maybe `cloud_management`? There are already multiple modules within starting with `resource`

supersergiy (approved):
💯

PR #405, opened 2023-05-20
fix: smaller integration tests with coverage, procs with subchunkable

dodamih:


dodamih:
Doesn't accomplish 100% coverage yet for subchunkable, but does fix the issues with the procs.

dodamih:
Should have 100% coverage now.

PR #408, opened 2023-05-26
feat: adjust masks and affinities for segmentation

macrintr:
These are some functions and specs that I used recently to adjust masks and affinities ahead of watershed and agglomeration. They will likely be useful for other segmentations. See commit messages and docstrings for more details. If these look reasonable, I'll add tests.

### DIFF
--- a/zetta_utils/mazepa_layer_processing/segmentation/masks/affinities.py
+++ b/zetta_utils/mazepa_layer_processing/segmentation/masks/affinities.py
@@ -1,0 +1,165 @@
+from __future__ import annotations
+
+from typing import Sequence
+
+import attrs
+import torch
+
+from zetta_utils import builder, mazepa, tensor_ops
+from zetta_utils.geometry.vec import Vec3D
+from zetta_utils.layer.volumetric.index import VolumetricIndex
+from zetta_utils.layer.volumetric.layer import VolumetricLayer
+
+
+@builder.register("AdjustAffinitiesOp")
+@mazepa.taskable_operation_cls
+@attrs.frozen
+class AdjustAffinitiesOp:
+    crop_pad: Sequence[int] = (0, 0, 0)
+
+    def get_input_resolution(  # pylint: disable=no-self-use
+        self, dst_resolution: Vec3D[float]
+    ) -> Vec3D[float]:
+        return dst_resolution
+
+    def with_added_crop_pad(self, crop_pad: Vec3D[int]) -> AdjustAffinitiesOp:
+        return attrs.evolve(self, crop_pad=Vec3D[int](*self.crop_pad) + crop_pad)
+
+    def __call__(
+        self,
+        idx: VolumetricIndex,
+        dst: VolumetricLayer,
+        aff_layer: VolumetricLayer,
+        aff_backup_layer: VolumetricLayer,
+        blackout_mask_layer: VolumetricLayer,
+        snap_mask_layer: VolumetricLayer,
+        threshold_mask_layer: VolumetricLayer,
+        threshold_value: float = 0.85,
+        fill_value: float = 0,
+        rework_mode: bool = False,
+        rework_mask_layer: VolumetricLayer | None = None,
+    ) -> None:
+        """
+        Adjust affinities using masks, making sure to backup any adjusted affinities.
+
+        :param idx: VolumetricIndex for the operation
+        :param src: layer with affinities
+        :param dst: layer where adjusted affinities will be written sparsely
+        :param aff_backup_layer: layer where original affinities will be stored
+            if chunk is adjusted
+        :param aff_backup_layer: layer where original affinities will be stored
+            if chunk is adjusted
+        :param blackout_mask_layer: one-hot mask to black out all affinities
+        :param snap_mask_layer: one-hot mask for which segments cannot span its
+            boundary ("snaps" objects at its boundary)
+        :param threshold_mask_layer: one-hot mask which will adjust affinities below
+            `threshold_value`
+        :param threshold_value: value for thresholding
+        :param fill_value: value to set the adjusted affinities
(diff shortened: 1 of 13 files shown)


zetta_utils/mazepa_layer_processing/segmentation/detect_consecutive_masks.py
```diff
@@ -0,0 +1,37 @@
...
+    of a mask. Using `num_consecutive=3`, this is identical to the boolean function:
+
+    `(z-2 & z-1 & z) | (z-1 & z & z+1) | (z & z+1 & z+2)`
+
+    :param src: tensor
+    :param num_consecutive: number z sections that a location must
+        be masked to be considered part of the consecutive mask
+
+    """
+    assert src.shape[-1] >= num_consecutive
```
dodamih, line 22: This should be an exception.

zetta_utils/mazepa_layer_processing/segmentation/adjust_affinities_with_masks.py
```diff
@@ -0,0 +1,109 @@
...
+    masks: Iterable[torch.Tensor],
+) -> torch.Tensor:
+    """
+    If any mask is present, return the data, otherwise return zeros
+
+    :param src: input Tensor
+    :param masks: list of masks
+    """
+    is_masked = any(mask.sum() > 0 for mask in masks)
+    has_data = src.sum() > 0
```
dodamih, line 20: This doesn't account for cases where the data is signed - there should be check to see whether the `src` is indeed unsigned.

zetta_utils/mazepa_layer_processing/segmentation/detect_consecutive_masks.py
dodamih, line 1: Do you think it might be better to rename this to `masks.py`, and refactor the mask boundary function from `adjust_affinities_across_mask_boundary` into it?

zetta_utils/mazepa_layer_processing/segmentation/adjust_affinities_with_masks.py
```diff
@@ -0,0 +1,109 @@
...
+    :param fill_value: value to set the adjusted affinities
+    """
+    mask_compare = mask[0, 1:, :, :] != mask[0, :-1, :, :]
+    aff_mask = torch.nn.functional.pad(mask_compare, (0, 0, 0, 0, 1, 0))
+    result[0, aff_mask] = fill_value
+    mask_compare = mask[0, :, 1:, :] != mask[0, :, :-1, :]
+    aff_mask = torch.nn.functional.pad(mask_compare, (0, 0, 1, 0, 0, 0))
+    result[1, aff_mask] = fill_value
+    mask_compare = mask[0, :, :, 1:] != mask[0, :, :, :-1]
+    aff_mask = torch.nn.functional.pad(mask_compare, (1, 0, 0, 0, 0, 0))
```
dodamih, line 84: See comment below, but it might be better to factor this part out into `masks.py`

zetta_utils/mazepa_layer_processing/segmentation/adjust_affinities_with_masks.py
dodamih, line 1: This file should be called `affinities.py` - part of me wonders if `adjust_affinities_with_masks` would be better off as an Operation (if the function is only going to be called in subchunkable and not by another function or an Operation, and does a lot of things all at once, then it's better off as an Operation), but it's a little borderline, especially since I can see this being used by a segmentation Operation in the future.

zetta_utils/mazepa_layer_processing/segmentation/detect_consecutive_masks.py
```diff
@@ -0,0 +1,37 @@
...
+    kernel = torch.ones((1, 1, num_consecutive), dtype=src.dtype)
+    # use full padding to help with propagation later
+    src_sum = torch.nn.functional.conv1d(reshaped_src, kernel, padding=num_consecutive - 1)
+    is_consecutive = src_sum >= num_consecutive
+    # propagate is_consecutive mask to all participating sections
+    participation = torch.zeros((x * y, c, z), dtype=torch.uint8)
+    for i in range(num_consecutive):
+        participation += is_consecutive[:, :, i : z + i]
+    # only needs to participate in one consecutive range to be masked
+    participates = participation > 1
```
dodamih, line 36: This might overflow in some weird cases. More concise to use the one-liner
`participation = torch.logical_or(*(is_consecutive[:, :, i : z+i] for i in range(num_consecutive))`

macrintr:
Adding the methods and specs I used to adjust Lee fly CNS affinities.

zetta_utils/mazepa_layer_processing/segmentation/masks/affinities.py
```diff
@@ -0,0 +1,163 @@
...
+        :param threshold_value: value for thresholding
+        :param fill_value: value to set the adjusted affinities
+        :param rework_mode: bool to use rework_mask_layer
+        :param rework_mask_layer: one-hot mask of locations that need to be reworked
+        """
+        idx_padded = idx.padded(self.crop_pad)
+        if rework_mode and (rework_mask_layer is not None):
+            rework_mask = rework_mask_layer[idx_padded]
+            if (rework_mask != 0).sum().item() == 0:
+                return
```
dodamih, line 66: Should handle case where ``rework_mode`` is set and ``rework_mask_layer`` is ``None``.

macrintr: If `rework_mode=True` but there's no `rework_mask_layer`, then don't you think it should just proceed as normal? Effectively, that means the `rework_mask_layer` is one-hot everywhere.

dodamih: That makes sense.

PR #409, opened 2023-05-26
feat: add skip_black_uploads check to CVBackend.write

macrintr:
This update provides the ability to sparsely overwrite an existing CloudVolume without deleting any data (requires `CloudVolume.delete_black_uploads=False`). It may provide a slight speedup if sparsely writing to a new CloudVolume (using `CloudVolume.delete_black_uploads=True`), because it will skip issuing `DELETE` requests to the storage backend. If this looks good, I'll write tests.

### DIFF
--- a/zetta_utils/layer/volumetric/cloudvol/backend.py
+++ b/zetta_utils/layer/volumetric/cloudvol/backend.py
@@ -178,19 +178,19 @@ class CVBackend(VolumetricBackend):  # pylint: disable=too-few-public-methods
         # Data in: cxyz
         # Write format: xyzc (b == 1)
 
-        data_np = tensor_ops.convert.to_np(data)
-        if data_np.size == 1 and len(data_np.shape) == 1:
-            data_final = data_np[0]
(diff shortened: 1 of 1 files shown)


dodamih:
Few things:
1. Since this requires `delete_black_uploads = False`, it should check the `cv_kwargs` to make sure that this flag is set accordingly during construction and raise a `ValueError`.
2. Sum might be 0 without the data actually being 0. Very unlikely, but possible for small chunks with +1, 0, -1 or something like that.
3. This seems like something that should be handled by CloudVolume with a cv_kwarg - CloudVolume already is checking for whether the data is 0, so shouldn't it handle not issuing requests as well?

macrintr:
@supersergiy suggested that if I want control of when to read & write from a layer, I could write my function as an `Op` ([example](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/mazepa_layer_processing/alignment/aced_relaxation_flow.py#L48)).

PR #414, opened 2023-06-09
fix?

nkemnitz:


zetta_utils/mazepa/execution_queue.py
dodamih, line 1: What's wrong with ``Iterable``? Wouldn't we want to use generators in some cases?

dodamih (approved):
I'm approving since I'm sure you had a good reason, but what's wrong with Iterable? Wouldn't we want to use generators in some cases (or was this what was causing flakey test issue)?

nkemnitz:
Hm, I think I misunderstood how Protocol's differ from inheritance. mypy was complaining, because mazepa.run_worker expects an ExecutionQueue, and not an SQSExecutionQueue

dodamih (approved):

PR #415, opened 2023-06-13
feat: allow torch functions in lambdas

macrintr:


nkemnitz:
If the other approach with the registered pytorch methods doesn't fit your use case, I am OK with this.

nkemnitz (approved):

PR #416, opened 2023-06-21
fix: gc cluster handling

akhileshh:


zetta_utils/mazepa_addons/execution_tracker.py
```diff
@@ -51,7 +51,10 @@ def register_execution(execution_id: str, clusters: list[ClusterInfo]) -> None:
 def read_execution_clusters(execution_id: str) -> list[ClusterInfo]:  # pragma: no cover
     row_key = execution_id
     col_keys = ("clusters",)
-    clusters_str = EXECUTION_DB[(row_key, col_keys)][col_keys[0]]
+    try:
+        clusters_str = EXECUTION_DB[(row_key, col_keys)][col_keys[0]]
+    except KeyError:
+        return []
```
supersergiy, line 57: Is not finding any clusters ever expected? Is it bad if we do not catch this error?

akhileshh: I saw an entry without the `clusters` column, not sure how that came to be. If we don't catch errors gc will keep failing and we will end up with stale entries. And we'd need to keep an eye on gc failures.

Do we need some notification system to alert when this happens?

supersergiy: Got it. Would be great to have a notification system to give us insight into how GC is doing in the future.

PR #426, opened 2023-06-29
feat: auto_divisibility, expand_bbox_resolution, expand_bbox_backend

dodamih:
Note that the legacy `expand_bbox` is now called `expand_bbox_processing` - all the CUE files have been updated. Docstring will be part of the Docs PR.


supersergiy (approved):
Thanks for patching all the specs!

PR #428, opened 2023-06-30
feat: kornia binary morph ops

nkemnitz:


nkemnitz:
@supersergiy Do you want tests here or is that more of a delegation thing?
Also, any reason to keep the custom `binary_closing` around?

supersergiy (commented):
This looks really cool!

Although the new functions don't have `if`/`for` statements, they're reasonably complex and would benefit from testing. The upside is that because there's no `if`'s, each function will need only a single test case to be fully covered.

Re deprecation: 
The `kornia_closing` requires kernel to be passed in explicitly, while the "custom" `binary_closing` just assumes something reasonable. The explicit kernel is useful in some scenarios, but making it a required argument will make it a hassle a lot of the time. I think ideally we'd make the kernel an optional argument, which would let us deprecate the non-kornia alternatives. 

Also, the `kernerl` argument should probably be typed, unless there's something I'm missing.

nkemnitz:
Set default kernel to square footprint of width=3, removed unnecessary conversion to floating point, added documentation

PR #430, opened 2023-06-30
fix: apply crop_pad after last write_proc

### LINKED ISSUE #429: Assymetric behavior between `processing_crop_pads` and rearrange in read/write procs
I want to run a morphological opening operation on a volume along in z. The operation however only works in 2D, and assumes the data is in `Z C X Y` shape, hence I added a read and write_proc:

```cue
processing_chunk_sizes: [[1280 * 2, 1024 * 2, 20]]
processing_crop_pads: [[0, 0, 4]]		// x y z
...
read_procs: [
	{
		"@mode":   "partial"
		"@type":   "rearrange"
		"pattern": "c x y z -> x y z c"	// Want to operate on yz axis (xz would also work)
	},
]
...
write_procs: [
	{
		"@mode":   "partial"
		"@type":   "rearrange"
		"pattern": "x y z c -> c x y z"
	}
]
```

That doesn't work because the padding of `processing_crop_pads` is applied before the first `read_proc` `rearrange` (which is fine), but the corresponding cropping is also applied before the `write_proc` `rearrange`, causing it to crop along the wrong axis.
Is there a scenario where those two operations should _not_ occur symmetrically? Otherwise I would just make a PR that ensures cropping runs as very last write_operation.

nkemnitz:
Resolves #429

### DIFF
--- a/zetta_utils/mazepa_layer_processing/common/volumetric_callable_operation.py
+++ b/zetta_utils/mazepa_layer_processing/common/volumetric_callable_operation.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import copy
+from functools import partial
 from typing import Callable, Generic, Sequence, TypeVar
 
 import attrs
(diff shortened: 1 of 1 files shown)


dodamih (approved):
Might be _slightly_ slow, but I think this is probably the best way to do it.

PR #438, opened 2023-07-05
feat: scheduling speed enhancement / bug fix

dodamih:


zetta_utils/geometry/bbox.py
```diff
@@ -443,6 +443,14 @@ def aligned(self, other: BBox3D) -> tuple[bool, ...]:  # pragma: no cover
         x_start, x_stop, y_start, y_stop, z_start, z_stop order."""
         return tuple(s[i] == o[i] for s, o in zip(self.bounds, other.bounds) for i in range(2))
 
+    def contained_in(self: BBox3D, other: BBox3D) -> bool:  # pragma: no cover
```
supersergiy, line 446: This needs a test

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -347,20 +348,25 @@ def _get_temp_dst(
...
         op_kwargs: P.kwargs,
     ) -> List[mazepa.tasks.Task[R_co]]:
-        tasks = [
-            self.op.make_task(
-                idx=idx_chunk,
-                dst=dst,
-                **op_kwargs,
+        with multiprocessing.Pool() as pool_obj:
+            tasks = pool_obj.map(
+                self._make_task,
```
supersergiy, line 367: Why is the `self._make_task` necessary here? Can we just map `self.op.make_task`?

supersergiy (changes requested):

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
dodamih: Cannot because parallel map requires a single argument.

supersergiy: Gotcha. In that case it's fine for the helper to be not typed tightly

supersergiy (approved):
Looks great!

PR #439, opened 2023-07-05
feat: `translation_granularity` for field-adjusted download

supersergiy:
Necessary for when field's `data_resolution` is coarser than the src `data_resolution`

### DIFF
--- a/zetta_utils/mazepa_layer_processing/alignment/common.py
+++ b/zetta_utils/mazepa_layer_processing/alignment/common.py
@@ -38,7 +38,7 @@ def translation_adjusted_download(
         xy_translation = cast(
             tuple[int, int],
             tuple(
-                translation_granularity * (e // translation_granularity)
+                translation_granularity * round(e / translation_granularity)
                 for e in xy_translation_raw
             ),
         )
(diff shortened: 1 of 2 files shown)


zetta_utils/mazepa_layer_processing/alignment/common.py
```diff
@@ -29,10 +30,18 @@ def translation_adjusted_download(
...
+    translation_granularity: int = 1,
 ) -> tuple[torch.Tensor, torch.Tensor | None, tuple[int, int]]:
     if field is not None:
         field_data = field[idx]
-        xy_translation = alignment.field.profile_field2d_percentile(field_data)
+        xy_translation_raw = alignment.field.profile_field2d_percentile(field_data)
+        xy_translation = cast(
+            tuple[int, int],
+            tuple(
+                translation_granularity * (e // translation_granularity)
```
nkemnitz, line 41: Probably doesn't make a big difference, but more accurate would be
```suggestion
                translation_granularity * round(e / translation_granularity)
```

nkemnitz (changes requested):
It still feels like something that we should be able to automatically determine, but I can't think of a good way to do it, at the moment. My old approach was filtering through the `read_procs` looking for any indication of lowest shared resolutions. That felt super hacky, though... I think leaving the responsibility to the user is OK for now.

Marked one rounding issue that could be slightly improved. Otherwise looks good.

nkemnitz (approved):

PR #442, opened 2023-07-10
feat: Skip intermediaries

dodamih:


supersergiy (commented):
Does there need to be a test where the output differs because intermediaries are skipped?

supersergiy (approved):

PR #450, opened 2023-07-30
fix(typing): Typeguard exceptions with X | Y union type under Python3.9

nkemnitz:
`from __future__ import annotations`  helps for static type checkers, but runtime checks will fail, e.g.:

```
│ zetta/zetta_utils/zetta_utils/builder/built_in_registrations.py:23 in                            │
│ efficient_parse_lambda_str                                                                       │
│                                                                                                  │
│ zetta/zetta_utils/venv/lib/python3.9/site-packages/typeguard/_decorators.py:140                  │
│ in wrapper                                                                                       │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: isinstance() arg 2 must be a type or tuple of types
```

### DIFF
--- a/zetta_utils/builder/build.py
+++ b/zetta_utils/builder/build.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import json
-from typing import Any, Callable, Final
+from typing import Any, Callable, Final, Optional
 
 import attrs
 from typeguard import typechecked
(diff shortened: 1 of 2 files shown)


supersergiy (approved):
Ah gee, this problem is back.

We have fixed this previously on the typeguard side (https://github.com/agronholm/typeguard/issues/283), but the solution changed several times and I'm guessing the pinned version of typeguard has some holes. I tried updating typeguard to latest and some other issues popped up. We can merge this for now, and I will follow up with `typeguard` -- we really want the users to just use PEP604 type union without worrying about it.

PR #451, opened 2023-07-30
Fixing some rounding related issues

nkemnitz:
2344e2b feels particularly ugly...


supersergiy (approved):
Great work, thanks!
Although some of this might not be particularly pretty, I do think it is the best we can do right now.

PR #456, opened 2023-08-01
Mypy bump

supersergiy:


zetta_utils/__init__.py
dodamih, line 1: Why does init need union? Also, why commented line?

zetta_utils/geometry/bbox.py
```diff
@@ -428,7 +428,7 @@ def pformat(self, resolution: Optional[Sequence[float]] = None) -> str:  # pragm
             f"({s.join([str(slice.stop) for slice in slices])})"
         )
 
-    def get_size(self) -> int | float:  # pragma: no cover
+    def get_size(self) -> Union[int, float]:  # pragma: no cover
```
dodamih, line 431: Is this change for readability or is there something broken with | syntax?

zetta_utils/layer/layer_set/backend.py
```diff
@@ -5,6 +5,8 @@
 
 import attrs
 
+from zetta_utils.layer.backend_base import Backend
```
dodamih, line 8: Double import with line 10?

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -184,17 +183,18 @@ def set_allow_cache(*args, **kwargs):
             and not arg.backend.is_local
             and not arg.backend.allow_cache
         ):
-            newarg = attrs.evolve(arg, backend=arg.backend.with_changes(allow_cache=True))
+            newv = arg.with_changes(backend=arg.backend.with_changes(allow_cache=True))
```
dodamih, line 186: This should be `newarg`.

zetta_utils/geometry/bbox.py
supersergiy: This is actually what was breaking the integration tests. Some typeguard dependency must have updated, so it started breaking on this even though we have its version pinned

PR #458, opened 2023-08-01
mazepa queueing refactor

supersergiy:
Previously, when a task got pushed to an ExecutionQueue, the ExecutionQueue would set `outcome_queue` attribute for the outgoing task. If a worker encountered an exception during the execution of the task, the exception would be propagated to the manager node through that `outcome_queue`. However, if a task failed to deserialize for whatever reason, the worker wouldn't know where to propagate the exception, and the manager node would keep hanging. 

In addition to such worker failures being silent, the old design relied on `callback` mechanisms to implement task completion reporting, task acknowledgment, and task lease extension. More specifically, the ExecutionQueue would attach several callbacks to each task, and those callbacks would be executed automatically upon a successful completion of the task. Although powerful and expressive, this callback based approach is not transparent, makes code hard to read and reason about. 

This PR breaks the concept of ExecutionQueue into the more natural Task+Outcome queue pair. It introduces a MessageQueue abstract class, which is just a lightweight interface that allows to push, pull, and acknowledge receipt of objects. The bulk of the logic that used to be done through task callbacks has now been moved to `worker.py`, making it more intuitive and readable. 

Lastly, the `TransientError` handling was also simplified -- a common set of transient error conditions is now is specified in `worker.py` and applied to every task. We can revive the custom per-task error conditions if the need arises, but I thought we should keep it simple until that happens.


nkemnitz (changes requested):
Do you know if/how this will impact performance?

I used it with my current alignment script and marked the two errors I ran into. Have you tried the big workflow tests Dodam added?

zetta_utils/cloud_management/resource_allocation/k8s/common.py
```diff
@@ -58,14 +58,15 @@ def _get_init_container_command() -> str:
     return command
 
 
-def get_worker_command(queue_spec: Dict[str, Any]):
+def get_mazepa_worker_command(task_queue_spec: dict[str, Any], outcome_queue_spec: dict[str, Any]):
     result = (
         """
     zetta -vv -l try run -s '{
         "@type": "mazepa.run_worker"
         exec_queue:
```
nkemnitz, line 66: `exec_queue` is gone
```suggestion
```

supersergiy:
Thanks @nkemnitz! 
I did try integration tests, but they don't go into cloud resource management.  \None of our CI tests that currently, which we should probably fixe in the future. I've actually had some of the bugs fixed but didn't commit it...

As to the performance, it should be pretty comparable. The only difference is the slightly slower acknowledgement of the outcome messages by the manager node, which I didn't optimize because I think it shouldn't be necessary with our tasks being large thanks to subchunkable. If we run into the issue of acks being too slow, we can optimize further pretty easily.

zetta_utils/mazepa/task_router.py
```diff
@@ -0,0 +1,37 @@
...
+    def push(self, payloads: Iterable[Task]):
+        tasks_for_queue = defaultdict(list)
+
+        for task in payloads:
+            matching_queue_names = [
+                queue.name for queue in self.queues if all(tag in queue.name for tag in task.tags)
+            ]
+            if len(matching_queue_names) == 0:
+                raise RuntimeError(
+                    f"No queue from set {list(self.queues)} matches " f"all tags {task.tags}."
```
dodamih, line 32: Should be 1 f-string.

dodamih (changes requested):
Definitely a cleaner design!

tests/unit/mazepa/test_execution_state.py
```diff
@@ -246,6 +246,27 @@ def test_task_outcome_fail_raise():
...
+    # type: () -> None
+    task = make_test_task(fn=lambda: None, id_="a")
+    flows = [make_test_flow(fn=dummy_iter, iterable=[task], id_="flow_0")]
+    state = InMemoryExecutionState(ongoing_flows=flows, raise_on_failed_task=True)
+    state.get_task_batch()
+    outcomes = {"i'm not from this flow": TaskOutcome[Any](exception=Exception())}
+    state.update_with_task_outcomes(outcomes)
+
+
+def test_task_outcome_fail_raise_unkown_task():
```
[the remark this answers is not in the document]
supersergiy: Too many n's!

supersergiy:
Integration tests: https://github.com/ZettaAI/zetta_utils/actions/runs/5751840821

dodamih (approved):

nkemnitz (changes requested):
Error propagation to the scheduler is not working at the moment. I'll try to come up with a small example to better investigate

Running job with main branch returns
```python
mazepa /home/nkemnitz/zetta/zetta_utils/zetta_utils/mazepa/execution_state.py: 133                       
Task traceback: Traceback (most recent call last):                      
  File "/opt/zetta_utils/zetta_utils/mazepa/tasks.py", line 194, in _call_without_upkeep                 
    return_value = self.fn(*self.args, **self.kwargs)                   
  File "/opt/zetta_utils/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py", line 73,
in __call__                            
    mazepa.Executor(do_dryrun_estimation=False, show_progress=False)(   
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 41, in __call__                          
    return execute(                    
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 115, in execute                          
    _execute_from_state(               
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 162, in _execute_from_state              
    process_ready_tasks(exec_queue, state, execution_id, max_batch_len) 
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 192, in process_ready_tasks              
    task_outcomes = queue.pull_task_outcomes()                          
  File "/opt/zetta_utils/zetta_utils/mazepa/execution_queue.py", line 64, in pull_task_outcomes          
    raise task.outcome.exception  # pragma: no cover                    
  File "/opt/zetta_utils/zetta_utils/mazepa/tasks.py", line 194, in _call_without_upkeep                 
    return_value = self.fn(*self.args, **self.kwargs)                   
  File "/opt/zetta_utils/zetta_utils/mazepa_layer_processing/alignment/compute_field_flow.py", line 84,  
in __call__                            
    src_data, src_field_data, src_translation = translation_adjusted_download(                           
  File "/opt/zetta_utils/zetta_utils/mazepa_layer_processing/alignment/common.py", line 57, in           
translation_adjusted_download          
    src_data = src[idx]                
  File "/opt/zetta_utils/zetta_utils/layer/volumetric/layer.py", line 33, in __getitem__                 
    return self.read_with_procs(idx=idx_backend)                        
  File "/opt/zetta_utils/zetta_utils/layer/layer_base.py", line 44, in read_with_procs                   
    data_backend = self.backend.read(idx=idx_proced)                    
  File "/opt/zetta_utils/zetta_utils/layer/volumetric/cloudvol/backend.py", line 200, in read            
    data_raw = cvol[idx.to_slices()]   
  File "/opt/zetta_utils/zetta_utils/layer/volumetric/index.py", line 46, in to_slices                   
    return self.bbox.to_slices(self.resolution, self.allow_slice_rounding)                               
  File "/opt/zetta_utils/zetta_utils/geometry/bbox.py", line 184, in to_slices                           
    tuple(                             
  File "/opt/zetta_utils/zetta_utils/geometry/bbox.py", line 185, in <genexpr>                           
    self.get_slice(i, resolution[i], allow_slice_rounding, round_to_int)
  File "/opt/zetta_utils/zetta_utils/geometry/bbox.py", line 149, in get_slice                           
    raise ValueError(                  
ValueError: BBox3D(bounds=((-2483.2, 81945.59999999999), (633216.0, 717644.6), (45.0, 90.0)), unit='nm', 
pprint_px_resolution=(1, 1, 1)) results in slice_end == 18495.9948453608 along dimension 1 at resolution 
== 38.8 while `allow_slice_rounding` == False.
```

whereas the PR simply returns
```python
mazepa /home/nkemnitz/zetta/zetta_utils/zetta_utils/mazepa/execution_state.py: 138                 
Task traceback: Traceback (most recent call last):                
  File "/opt/zetta_utils/zetta_utils/mazepa/tasks.py", line 84, in __call__                        
    return_value = self.fn(*self.args, **self.kwargs)             
  File "/opt/zetta_utils/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py", line 73, in __call__               
    mazepa.Executor(do_dryrun_estimation=False, show_progress=False)(                              
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 47, in __call__                    
    return execute(              
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 127, in execute                    
    _execute_from_state(         
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 176, in _execute_from_state        
    submit_ready_tasks(task_queue, outcome_queue, state, execution_id, max_batch_len)              
  File "/opt/zetta_utils/zetta_utils/mazepa/execution.py", line 216, in submit_ready_tasks         
    state.update_with_task_outcomes(                              
  File "/opt/zetta_utils/zetta_utils/mazepa/execution_state.py", line 139, in update_with_task_outcomes                             
    raise MazepaExecutionFailure("Task failure.")                 
zetta_utils.mazepa.exceptions.MazepaExecutionFailure: Task failure.

**Comment by nkemnitz** [comment]
To replicate. The info for the source layer is missing a `key`, which causes the lowest-level chunks on the remote worker to fail upon creating the CloudVolume.
That part works fine - the task outcomes contain the correct Exception and traceback, and I am pretty sure the remote scheduler also correctly [logs this as error message](https://github.com/ZettaAI/zetta_utils/blob/sergiy/mazepa_queing_refactor/zetta_utils/mazepa/execution_state.py#L138). However, it's doing so on the remote worker/scheduler, and the `MazepaExecutionFailure` that follows will generate and forward only its own stacktrace to user.

```cue
target: {
    "@type": "build_subchunkable_apply_flow"
    bbox: {
        "@type": "BBox3D.from_coords"
        start_coord: [0, 0, 0]
        end_coord: [4096, 4096, 1]
        resolution: [4, 4, 40]
    }
    dst: {
        "@type": "build_cv_layer"
        path: "gs://tmp_2w/nkem/debug_subchunkable2"
        info_reference_path: "gs://tmp_2w/nkem/debug_subchunkable"
    }
    dst_resolution: [4, 4, 40]
    processing_chunk_sizes: [[4096, 4096, 1], [1024, 1024, 1]]
    processing_crop_pads: [[0, 0, 0], [0, 0, 0]]
    skip_intermediaries: true
    fn: {
        "@type":      "lambda"
        "lambda_str": "lambda src: src"
    }
    op_kwargs: {
        src: {
            "@type": "build_cv_layer"
            path: "gs://tmp_2w/nkem/debug_subchunkable"
            cv_kwargs: {"fill_missing": false}  // To cause exception
        }
    }
}

"@type":                "mazepa.execute_on_gcp_with_sqs"
[...]
worker_replicas: 1
local_test:      false
```

nkemnitz (approved):
Nice job, thanks for your patience! Assuming the tests will pass, I am OK with merging, too.

PR #460, opened 2023-08-03
hotfix: Let skip_intermediaries through with crop_pad at the bottom, update to use expand_bbox_processing, and update old specs

dodamih:
3 lines other than spec updates.


supersergiy (approved):
Nice work!

PR #461, opened 2023-08-03
fix(compute_field_flow): skip intermediaries if possible

nkemnitz:
Forgot to mention this one to @dodamih before #460 

Although, glancing at the current logic (see below), I believe `skip_intermediaries` always must be `not self.intermediaries_dirs` - maybe we can just get rid off it?

https://github.com/ZettaAI/zetta_utils/blob/59e53681c946bd335b3b46c9aff283a3b5d33986/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py#L251-L276


dodamih:
I don't have a good handle on how ComputeFieldFlow is used, but this is okay as far as Subchunkable is concerned (I wouldn't do the same thing is subchunkable because automatically skipping intermediaries seems dangerous though).

dodamih (approved):

PR #463, opened 2023-08-04
Convblock changes from Kisuk

dodamih:
Ignoring the JIT case for testing.

### DIFF
--- a/zetta_utils/convnet/architecture/convblock.py
+++ b/zetta_utils/convnet/architecture/convblock.py
@@ -151,7 +151,7 @@ class ConvBlock(nn.Module):
         for i, layer in enumerate(self.layers):
             if (i in self.skips_dst) and (conv_count in skip_data_for):
                 # In tracing mode, shapes obtained from tensor.shape are traced as tensors
-                if isinstance(result.shape[0], torch.Tensor):  # type: ignore
+                if isinstance(result.shape[0], torch.Tensor):  # type: ignore # pragma: no cover
                     size = list(map(lambda x: x.item(), result.shape))  # type: ignore
                 else:
                     size = result.shape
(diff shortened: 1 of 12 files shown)


nkemnitz (commented):
You are potentially changing the order of layers in the UNet which may cause incompatibilities with existing trained nets. Need to bump this version to 0.0.2 and move 0.0.1 to the deprecated directory.

dodamih:
@nkemnitz in `main`, there isn't any deprecated UNets - isn't this correct? (What's in main is what we have in the PR as 0.0.0).

nkemnitz:
[0.0.0](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/convnet/architecture/deprecated/unet.py) is deprecated. Most recent one is [0.0.1](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/convnet/architecture/unet.py)

zetta_utils/convnet/architecture/convblock.py
```diff
@@ -9,13 +9,15 @@
...
+from zetta_utils.tensor_ops import crop_center
 from zetta_utils.typing import ensure_seq_of_seq
 
 Padding: TypeAlias = Union[Literal["same", "valid"], Sequence[int]]
 PaddingMode: TypeAlias = Literal["zeros", "reflect", "replicate", "circular"]
+ActivationMode: TypeAlias = Literal["pre", "post"]
 
 
-@builder.register("ConvBlock")
+@builder.register("ConvBlock", versions=">=0.0.1")
```
supersergiy, line 20: We're trying to keep the versions tied to `zetta_utils` release version in the `pyproject.toml`.
Because it's a breaking change, we would need to bump the version to 0.0.2, and I think that's the version that the new ConvBlock should require.

supersergiy (changes requested):
All the code changes look good to me. The only issue is the versioning -- as I mentioned in the comments, because it's a breaking change we'd want to bump the version in pyproject.toml, and use that version when registering the new primitives.

zetta_utils/convnet/architecture/deprecated/convblock.py
```diff
@@ -0,0 +1,139 @@
...
+from typing_extensions import TypeAlias
+
+from zetta_utils import builder
+from zetta_utils.typing import ensure_seq_of_seq
+
+Padding: TypeAlias = Union[Literal["same", "valid"], Sequence[int]]
+PaddingMode: TypeAlias = Literal["zeros", "reflect", "replicate", "circular"]
+
+
+@builder.register("ConvBlock", versions="==0.0.0")
```
supersergiy, line 18: And this would be `<=0.0.1`

PR #465, opened 2023-08-07
feat: Import Everything option for python-style specs (experimental)

supersergiy:
Python spec files, which are still in the testing phase, can now pull in all primitives from the entire repo with:

```python
from zetta_utils.everything import *
```

The `everything.py` file is autogenerated by looking at the repo's source code, as doing it manually would be too hard to maintain. The generator ensures there aren’t any naming conflicts. If it finds any, it'll throw an error. If python specs get more popular, we might want to add a check for duplicates in our CI.

`everything.py` doesn't include anything from `convnet.architecture.deprecated`, as we don't have any old python specs to deal with. Python specs don’t have the versioning capabilities that the builder does, so we need to think of something different. One idea is to have different versioned `everything` files. For example, `everything_v0.py` could import the old UNet, and `everything_v1.py` the new one. The spec version would be clear from the get-go, as in `from zetta_utils.everything_v0 import *`. If we like this idea, maybe I should rename the current `everything.py` to `everything_v0.py` to start with clear versioning.

This PR also has a couple more tweaks: importing from zetta now sets up the logger automatically, and I've done a bit of tidying to avoid naming issues.

### DIFF
--- a/scripts/generate_everything_import.py
+++ b/scripts/generate_everything_import.py
@@ -1,0 +1,94 @@
+from __future__ import annotations
+
+import ast
+import importlib
+import os
+import pkgutil
+from types import ModuleType
+
+from zetta_utils import log
+
+logger = log.get_logger("zetta_utils")
+
+
+class NameOverlapError(Exception):
+    """Custom exception for name overlaps in modules."""
+
+
+def get_defined_objects(filepath: str) -> list[str]:
+    with open(filepath, "r", encoding="utf-8") as f:
+        tree = ast.parse(f.read(), filename=filepath)
+        return [
+            node.name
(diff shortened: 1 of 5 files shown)


supersergiy:
Updated `from zetta_utils.everything import *` to `from zetta_utils.api.v0 import *` to enable versioning.

Made the logger configurable only on `api` imports, not an all zutils imports as it configures rich log handler which could be invasive in some environments.

nkemnitz (approved):
Thanks! The changes shouldn't interfere with existing code or workflows, so it's probably safe to merge and let users try if this is more comfortable to work with than the cue specs

PR #466, opened 2023-08-08
Rounding Issues... Round 2

nkemnitz:
Resolves most of the floating point comparison issues: there are the obvious comparisons (`==`, etc.), for which you can now use `allclose`, but `ceil` and `floor` were more frequently cause for trouble in our code.

Typical example is `floor(0.6/0.2) -> 2`. Resolving this for now with `floor(round(0.6/0.2, 10))`, but it does look a bit ugly...

### DIFF
--- a/zetta_utils/geometry/bbox_strider.py
+++ b/zetta_utils/geometry/bbox_strider.py
@@ -45,11 +45,11 @@ class BBoxStrider:
     chunk_size: Vec3D[int]
     stride: Vec3D[int]
     max_superchunk_size: Optional[Vec3D[int]] = None
-    stride_start_offset_in_unit: Optional[Vec3D[int]] = None
+    stride_start_offset_in_unit: Optional[Vec3D] = None
     chunk_size_in_unit: Vec3D = attrs.field(init=False)
     stride_in_unit: Vec3D = attrs.field(init=False)
     bbox_snapped: BBox3D = attrs.field(init=False)
-    step_limits: Tuple[int, int, int] = attrs.field(init=False)
+    step_limits: Vec3D[int] = attrs.field(init=False)
     step_start_partial: Tuple[bool, bool, bool] = attrs.field(init=False)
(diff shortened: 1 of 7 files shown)


zetta_utils/geometry/bbox_strider.py
dodamih, line 1: Not sure if we want to use Vec3D for non-number types. Also, can we use a variable instead of using hardcoding 10?

zetta_utils/geometry/vec.py
dodamih, line 1: Same thing about Vec3D[bool].

dodamih (changes requested):

zetta_utils/geometry/bbox_strider.py
nkemnitz: Sorry, which line / non-number type is this referring to?

supersergiy: I think @dodamih is worried about using `Vec3D[bool]` on lines 52, 53 etc

We're dynamically looking into the types to return a correct one after many [[arithmetic operations](https://github.com/ZettaAI/zetta_utils/blob/5514fea68ceb24d6e48c2bb212611ac18e6ae5ae/zetta_utils/geometry/vec.py#L187-L190)], and it's not covered in any of the tests. 

Not sure if we want to extend Vec3D to correctly do arithmetic with `bool`'s, or just throw an exception there.

dodamih: I'd push for exceptions - for one thing, `bool`s are not an Abelian group. There's also the added complexity from having to type infer correctly with bools, and the worry is that a user would see this and say "oh, this Vec3D container must support non-number types too".

nkemnitz:
Removed all traces of `Vec3D[bool]` and unified the precision related rounding calls to use a single constant `VEC3D_PRECISION`

supersergiy (approved):

dodamih (approved):

PR #471, opened 2023-08-09
feat: adds sqs region support

akhileshh:


supersergiy (approved):
Looks good. It's pretty clever to use the SQSQueue objects here

PR #490, opened 2023-08-17
feat: allow supplying no op_kwargs

dodamih:
`test_uint8_copy_bbox` was a missing integration test.


supersergiy (approved):
Thanks for the fix

PR #491, opened 2023-08-17
minor fixes

nkemnitz:
* `scripts/generate_everything_import.py`
  * skip import errors encountered during traversal
  * fix missing new lines at end
  * set default log level to `INFO`

* `make_ng_layer`
  * change `layers` type to `Sequence[Tuple[str, str, Optional[str]]]`
  * handle empty `source` entry (for layers of type `annotation`)


supersergiy (approved):
Thanks for the fixes, looks good!

PR #496, opened 2023-08-21
fix: running convnet on CPU

trivoldus28:
For whatever reason running torch on CPU requires the `map_location` to be specified

### DIFF
--- a/zetta_utils/convnet/utils.py
+++ b/zetta_utils/convnet/utils.py
@@ -64,7 +64,10 @@ def load_weights_file(
         return model
 
     with fsspec.open(ckpt_path) as f:
-        loaded_state_raw = torch.load(f)["state_dict"]
+        if torch.cuda.is_available():
+            loaded_state_raw = torch.load(f)["state_dict"]
+        else:
+            loaded_state_raw = torch.load(f, map_location=torch.device('cpu'))["state_dict"]
         if component_names is None:
             loaded_state = loaded_state_raw
         elif remove_component_prefix:


supersergiy:
Should this default to `cuda` when available, or accept an optional `device` arg?

trivoldus28:
@supersergiy reading the [doc](https://pytorch.org/docs/stable/generated/torch.load.html), seems like the default `None` will make `load` puts the model to wherever they were trained on regardless if there is actually a GPU.

 In this case I think using a `device` arg could work.

supersergiy:
@nkemnitz @dodamih do you guys have a preference here?

dodamih:
I think supporting `device` is better (don't think there's a need to support the other options in `map_device` though.)

nkemnitz:
If I recall correctly, we started using the `device = 'cuda' if torch.cuda.is_available() else 'cpu'` so that for enormous render jobs (or in times of GPU sparsity) you could add a CPU-only node pool to work on remaining tasks in the queue. Hardcoding `device='cuda'` in the task spec caused all CPU workers to crash.

We can still use the optional device argument, but should decide how to handle unfulfillable requests.

trivoldus28:
Maybe add `device = 'dynamic'` option? Which will be set to either cpu or cuda at run-time, reducing the need for using `device = 'cuda' if torch.cuda.is_available() else 'cpu'` outside of `convnet.utils`

trivoldus28:
@supersergiy haha you got me

supersergiy:
@trivoldus28 is there a plan to finish this or should we close?

trivoldus28:
Probably can close for now until it becomes a problem?

PR #497, opened 2023-08-21
fix: make get_affine_field() returns CXYZ tensors

trivoldus28:


supersergiy:
Looks like `pylint-isort` check is failing -- have you installed `pre-commit`? It's pretty helpful

trivoldus28:
Ah yea sorry let me try `pre-commit` - didn't think it would triggered it :sweat_smile: . Also I guess there are some unit tests need fixings too.

nkemnitz:
This change will require updates to existing regimes, e.g. the [BaseEncoder](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/training/lightning/regimes/alignment/base_encoder.py#L77C11-L85)

trivoldus28:
Good catch. And looking at existing usages it seems hard for me to do the updates

supersergiy:
There's 3 places where `get_affine_field` is used, and I believe adding a simple rearrange at each one should be sufficient?

Btw @trivoldus28 for your context:
You might have figured this out already, but training regimes and high-level layer processing constructs are exempt from unit test coverage requirement. This is done for 2 reasons. First, these highest-level constructs are often changing, and we do not want to inhibit the pace of research. Second, testing some of these high level constructs is technically challenging/slow. Details on this can be seen in the [[coveragerc](https://github.com/ZettaAI/zetta_utils/blob/main/.coveragerc)] file. We also have integration tests that cover `subchunkable` and `volumetric_apply` flows which are not a part of the usual CI, but can be run on demand through github actions. The exact testing strategy is still not finalized and is up for debate

trivoldus28:
Yea though a bit hacky I can def make a first pass of rearrangement.

supersergiy:
Interesting, why do you think it's hacky?

trivoldus28:
Actually taking a second look, maybe it's not so bad. In this [example](https://github.com/ZettaAI/zetta_utils/blob/b00c106c0fd102786d779c776991456a9c1356a7/zetta_utils/training/lightning/regimes/alignment/encoding_coarsener_gen_x1.py#L95-L102C19), do you mean something like from:
```
        equivar_field_inv = (~seed_field.from_pixels())(
            tensor_ops.transform.get_affine_field(
                size=src.shape[-1],
                rot_deg=-equivar_rot,
            )
            .field()  # type: ignore
            .to(src.device)
        ).pixels()
```
to
```
        equivar_field_inv = (~seed_field.from_pixels())(
            tensor_ops.transform.get_affine_field(
                size=src.shape[-1],
                rot_deg=-equivar_rot,
            )
            .movedim(-1, 0)  # CXYZ -> ZCXY
            .field()  # type: ignore
            .to(src.device)
        ).pixels()
```

supersergiy:
Yeah, that's pretty close!

You were disciplined and left a comment explaining the dimension change. We like to force such good behavior through always using einops rearrange function when messing with dims. You can check it out, it's sprinkled all over the codebase

supersergiy (approved):
Look great!

PR #498, opened 2023-08-22
Support mapping as input for most tensor ops

### LINKED ISSUE #447: A decorator that would make `tensor_op`'s work on `dict[str, Tensor]` input
Need `procs` to handle dicts for multi input/output processing.
Decorator approach will reduce code duplication and testing load

nkemnitz:
Should resolve #447.

Only needed it for `rearrange` and `crop` to work with my input/output layer set, though.

### DIFF
--- a/mypy_plugin.py
+++ b/mypy_plugin.py
@@ -3,9 +3,76 @@ from __future__ import annotations
 
 from typing import Callable, Final, Optional
 
-from mypy.plugin import ClassDefContext, Plugin
+from mypy.nodes import ARG_POS
+from mypy.plugin import ClassDefContext, FunctionContext, Plugin
 from mypy.plugins.common import add_method_to_class  # add_attribute_to_class,
-from mypy.types import AnyType, Parameters, TypeOfAny
+from mypy.types import (
+    AnyType,
+    CallableType,
+    Instance,
+    Overloaded,
+    Parameters,
+    TypeOfAny,
+    TypeVarType,
+)
+
+MAZEPA_INSTALLED = False
+try:
+    from zetta_utils import mazepa  # pylint: disable
+
+    MAZEPA_INSTALLED = True
+except ImportError:
+    pass
+
+
+def supports_dict_callback(ctx):
+    original_function_type = ctx.arg_types[0][0]
+
+    if not isinstance(original_function_type, CallableType):
+        ctx.api.fail("Argument to 'supports_dict' must be callable", ctx.context)
+        return original_function_type
+
+    if not original_function_type.arg_types:
+        ctx.api.fail("Function must have at least one argument", ctx.context)
+        return original_function_type
+
+    original_arg_type = original_function_type.arg_types[0]
+
+    if isinstance(original_arg_type, AnyType):
+        ctx.api.fail("The first argument must be annotated", ctx.context)
+        return original_function_type
+
+    # if (
+    #     original_arg_type.type.fullname == "builtins.dict"
+    #     or original_arg_type.type.fullname == "typing.Mapping"
+    # ):
+    #     ctx.api.fail(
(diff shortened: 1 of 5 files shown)


supersergiy:
Very interesting typing problem...
Not sure if we can use overload on a decorator like this to overload the decorated function. Maybe it's more proper to use type variable here to link returned function's input and output type?

nkemnitz:
Have you ever seen a similar pattern elsewhere? It looks a bit clumsy for something that the user could do in a one-liner like `{k: rearrange(v, ...) for k, v in layers.items()}` - just complicated with the cue specs.

nkemnitz:
>Maybe it's more proper to use type variable here to link returned function's input and output type?

I tried that at first, but you have functions (`to_np`) that are meant to change the type.

nkemnitz:
Think that last commit should fix the remaining issues (`Unexpected keyword argument "data" for "interpolate"`). That is a bit annoying, since type hints shouldn't require changes to function calls...

nkemnitz:
Oh, now I think I understand what you mean. `support_dict` always only takes a `Callable` that expects a `TensorTypeVar` as input, and the `wrapper` function inside is the overloaded one that returns a `Callable` whose output depends on whether the input is a mapping or not... I can let mypy do the painful work of calculating the return type... but I don't think we can actually annotate anything as `Overload` yet(?):

```python
def supports_dict(func: Callable[Concatenate[TensorTypeVar, P], ReturnTensorTypeVar]):
    @overload
    def wrapper(
        data: Mapping[HashableTypeVar, TensorTypeVar], *args: P.args, **kwargs: P.kwargs
    ) -> dict[HashableTypeVar, ReturnTensorTypeVar]:
        ...

    @overload
    def wrapper(data: TensorTypeVar, *args: P.args, **kwargs: P.kwargs) -> ReturnTensorTypeVar:
        ...

    def wrapper(data, *args, **kwargs):

        if isinstance(data, Mapping):
            return {k: func(v, *args, **kwargs) for k, v in data.items()}
        else:
            return func(data, *args, **kwargs)

    reveal_type(wrapper)
    return wrapper
```
```bash
Revealed type is "Overload(def [HashableTypeVar <: typing.Hashable] (data: typing.Mapping[HashableTypeVar`-4, TensorTypeVar`-1], *P.args, **P.kwargs) -> builtins.dict[HashableTypeVar`-4, ReturnTensorTypeVar`-3], def (data: TensorTypeVar`-1, *P.args, **P.kwargs) -> ReturnTensorTypeVar`-3)"
```

supersergiy:
How about just using `TypeVar`, and not using the decorator for `to_np`, `to_torch`, `as_type`?

nkemnitz:
>How about just using TypeVar, and not using the decorator for to_np, to_torch, as_type?
Fine with me, but it doesn't seem to cause harm.

Bigger issue now is that the typecheckers are completely happy with `convert.to_float32("Hello")` and I don't understand why.

supersergiy:
Does that happen if you remove the `overlaod` form the decorator?

nkemnitz:
Good idea. Still happens when I remove the overloads from `wrapper`. But works as expected if I additionally set the return type of `supports_dict`...
So likely mypy _needs_ the return type of the outer function... which I don't know how to provide with overload

supersergiy:
I didn't mean your new solution with `wrapper`, I mean the original decorator `supports_dict` -- fully annotate it with a type variable that can take on either Torch | NDArray or dict of them, and returns the same type

supersergiy:
Perhaps we should just meet tomorrow to discuss?

supersergiy:
@nkemnitz https://github.com/python/mypy/issues/16137 doesn't seem to be getting any attention, so I extended our mypy plugin to handle your `supports_dict` function. I think we're good to merge now. Sorry this took so long -- typing system bugs are always extra painful

supersergiy (approved):

PR #502, opened 2023-08-26
feat(training): Multinode DDP with Kubernetes

akhileshh:
Tried to make hybird k8s plus instance group training work but encountered too many network issues. Proceeding in that direction doesn't seem worth it since it is much easier with k8s.

Also removed default taint tolerations in `job` so gc cronjob can run on the system pool.


supersergiy (approved):
Looks solid!

PR #503, opened 2023-08-28
fix: taint based on num_nodes; return vars consistency

akhileshh:


supersergiy (approved):
Thanks for the fix and for leaving that comment in!

PR #504, opened 2023-08-30
fix(layer_set): missing with_changes

nkemnitz:


dodamih:
Oh, I was working on my own fix for this - I thought the better way to do this was to add it to the base layer.py and replace `with_procs`, so that is what I was writing tests for. Do you have thoughts?

nkemnitz:
No preference - just ran into the bug when writing layer sets and needed a fix.

dodamih:
I'll merge it in for now and make an issue for refactoring.

dodamih (approved):

PR #510, opened 2023-09-07
feat: pod_failure_policy to ignore disruption

nkemnitz:
Still needs testing on a cluster that supports Pod Failure Policies


akhileshh (approved):
looks good!

Issue #500, opened 2023-08-22
Precomputed specs to allow easy creation of new scales


supersergiy:
* A list of scales to be created would be provided as a separate argument
* Optional reference scale would default to MIP0. (@trivoldus28 I was wrong before, having a reference scale would be useful in many scenarios)
* `chunk_size`, `dataset_size`, and `voxel_offset` maps could be used to specify small deviations from the reference scale

Questions:
* What if both new scale list and `info_field_overrides: {'scales': ...}` are provided? Should an exception be thrown, or should the new scales be created on top of the initial scales provided by the overrides?

trivoldus28:
Thanks for making the issue @supersergiy . I'm gonna write up some draft examples of how the interface is going to be like and yea I think having a ref scale would be useful. I'll take a look of how `info_field_overrides` are being used first tho to cover most common use cases.

On handling `info_field_overrides` with `scales` specified, I think we should just throw an exception for now unless it's useful to merge it and adding scales.

nkemnitz:
`size`, `voxel_offset`, `key` are the keys that are directly linked to resolution. At least during our current alignment and segmentations steps, the shrink factor is always `[2,2,1]`. For the final render step, we sometimes downsample in z as well to keep voxel size isotropic. `chunk_sizes` often stays the same across scales.

Ideally we would check that for all the requested scales, the reference values make sense.
E.g. requesting 3 scales with a voxel_offset of 30 at scale 0 will cause issues with the visualization at scale 3, because the image suddenly jumps half a pixel

trivoldus28:
Example of what I think could work:
```

{
    "@type": "build_cv_layer"
    path: #DST_PATH
    ref_info: ...

    // add_scales_ref can either be a dictionary
    add_scales_ref: {
        "key": "4_4_40",
        "resolution": [4, 4, 40],
        "chunk_sizes": [32, 32, 1],
        "encoding": "raw",
        "size": #OUTPUT_SIZE,
        "voxel_offset": [0, 0, 0]
    }
    // Or an existing key from `ref_info` from which the dict will be copied from
    // Throws an error if `ref_info` is not provided
    add_scales_ref: "4_4_40"

    // Example adding 3 scales
    add_scales: [
        {
            // full specs for this scale
            "key": "8_8_80",
            "resolution": [8, 8, 40],
            "chunk_sizes": [32, 32, 1],
            "encoding": "raw",
            "size": #OUTPUT_SIZE,
[...]

PR #511, opened 2023-09-07
Initial implementation of info_add_scales

trivoldus28:
For issue https://github.com/ZettaAI/zetta_utils/issues/500.

Update 10/6/23: `add_scales_ref = None` (default) will use the highest resolution scale as reference.

Annotated example of current implementation:
```
{
    "@type": "build_cv_layer"
    path: #DST_PATH
    ref_info: ...

    // add_scales_ref can either be a dictionary
    add_scales_ref: {
        "key": "4_4_40",
        "resolution": [4, 4, 40],
        "chunk_sizes": [[512, 512, 1]],
        "encoding": "raw",
        "size": [1000, 1000, 1000],
        "voxel_offset": [0, 0, 0]
    }
    // Or an existing key from `ref_info` from which the dict will be copied from.
    // Throws an error if `ref_info` if key is not found
    add_scales_ref: [4, 4, 40]
    // Or `None` to use the highest resolution scale
    add_scales_ref: null

    // Example adding 3 scales.
    add_scales: [
        {
            // Example of a fully speced entry
            "key": "asdf8", // can be anything
            "resolution": [8, 8, 40],
            "chunk_sizes": [[32, 32, 1]],
            "encoding": "raw",
            "size": [1000, 1000, 1000],
            "voxel_offset": [1, 2, 3]
        },

        {   
            // Example of a partial entry - missing parameters will be filled in by scaling values from `add_scales_ref`
            // `resolution` is the only required parameter. `key` will be `resolution` underscore-separated if not given.
            // The following required properties will be computed if not specified: chunk_sizes, encoding, size, & voxel_offset
            // Error cases:
            // 1. `add_scales_ref` doesn't have values for required properties
            // 2. Values are not integers after scaling.
            // To resolve, just provide the mismatched properties here.
            "resolution": [16, 16, 40],
        },  

        [32, 32, 40],  // shorthand, equivalent to `{"resolution": [8, 8, 40]}`
    ]

    // two modes: "merge" (default) and "replace"
    info_add_scales_mode: "merge"       // "merge" will merge added scales to existing scales from `ref_info` (if provided)
    info_add_scales_mode: "replace"     // "replace" will not keep existing scales

    info_chunk_size: [1024, 1024, 1]  // using ``info_[chunk_size,dataset_size,voxel_offset][_map]`` will override the computed values
}
```

### DIFF
--- a/tests/unit/layer/volumetric/precomputed/test_precomputed.py
+++ b/tests/unit/layer/volumetric/precomputed/test_precomputed.py
@@ -23,6 +23,8 @@ LAYER_X6_PATH = "file://" + os.path.join(INFOS_DIR, "scratch", "layer_x6")
 LAYER_X7_PATH = "file://" + os.path.join(INFOS_DIR, "scratch", "layer_x7")
 NONEXISTENT_LAYER_PATH = "file://" + os.path.join(INFOS_DIR, "scratch", "nonexistent_layer")
 
+EXAMPLE_INFO = '{"data_type": "uint8", "num_channels": 1, "scales": [{"chunk_sizes": [[1024, 1024, 1]], "encoding": "raw", "key": "4_4_40", "resolution": [4, 4, 40], "size": [16384, 16384, 16384], "voxel_offset": [0, 0, 0]}, {"chunk_sizes": [[2048, 2048, 1]], "encoding": "raw", "key": "8_8_40", "resolution": [8, 8, 40], "size": [8192, 8192, 8192], "voxel_offset": [0, 0, 0]}], "type": "raw"}'
+
 _write_info_notmock = precomputed._write_info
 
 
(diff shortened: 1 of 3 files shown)


trivoldus28:
Guess I'll need to write some tests. Also will be test driving this feature a little bit more

trivoldus28:
Not sure why CI `mypy` gives these rather unrelated errors:
```
zetta_utils/tensor_ops/transform.py:39: error: Unused "type: ignore" comment  [unused-ignore]
zetta_utils/tensor_ops/common.py:20: error: Incompatible return value type (got "object", expected "Union[Tensor, ndarray[Any, dtype[Any]]]")  [return-value]
zetta_utils/tensor_ops/common.py:25: error: Incompatible return value type (got "object", expected "Union[Tensor, ndarray[Any, dtype[Any]]]")  [return-value]
Found 3 errors in 2 files (checked 280 source files)
Error: Process completed with exit code 1.
```
Maybe mypy got updated?

zetta_utils/layer/volumetric/cloudvol/build.py
```diff
@@ -34,6 +34,9 @@ def build_cv_layer(  # pylint: disable=too-many-locals
     info_dataset_size_map: dict[str, Sequence[int]] | None = None,
     info_voxel_offset: Sequence[int] | None = None,
     info_voxel_offset_map: dict[str, Sequence[int]] | None = None,
+    info_add_scales: Sequence[Sequence[int] | dict[str, Any]] | None = None,
+    info_add_scales_ref: str | dict[str, Any] | None = None,
+    info_add_scales_mode: str = "merge",
```
supersergiy, line 39: Can be typed more precisely as `Literal['merge', 'replace']`

zetta_utils/layer/volumetric/precomputed/precomputed.py
```diff
@@ -105,7 +183,25 @@ def make_info(  # pylint: disable=too-many-branches, consider-iterating-dictiona
             reference_info = {}  # type: Dict[str, Any]
             if self.reference_path is not None:
                 reference_info = get_info(self.reference_path)
-            result = deepcopy({**reference_info, **field_overrides})
+
+            result = deepcopy(reference_info)
+
+            if self.add_scales is not None:
+                if "scales" in field_overrides:
```
supersergiy, line 190: great check

zetta_utils/layer/volumetric/precomputed/precomputed.py
```diff
@@ -58,6 +59,76 @@ def _str(n: float) -> str:  # pragma: no cover
...
+    return ret
+
+
+def _merge_and_sort_scales(
+    existing_scales: Sequence[dict[str, Any]], new_scales: Sequence[dict[str, Any]]
+) -> Sequence[dict[str, Any]]:
+    """Merge two list of scales, overwriting the old with new entries"""
+    ret_dict = dict(zip([k["key"] for k in existing_scales], existing_scales))
+    ret_dict.update(dict(zip([k["key"] for k in new_scales], new_scales)))
+    # sort scales by voxel volumes
```
supersergiy, line 124: We should probably mention that the scales will get sorted in the docstring to `build` on the off chance that a user wants to do something cheeky and expects the scales to be given in the order provided by them

zetta_utils/layer/volumetric/precomputed/precomputed.py
```diff
@@ -58,6 +59,76 @@ def _str(n: float) -> str:  # pragma: no cover
     return str(n)
 
 
+def _check_seq_is_int(seq: Sequence[int | float]) -> bool:
+    error = False
+    for k in seq:
+        error |= not float(k).is_integer()
+    return error
```
supersergiy, line 66: A slightly more pythonic way to achieve this:
```python
return not all(float(k).is_integer() for k in seq)
```

zetta_utils/layer/volumetric/precomputed/precomputed.py
```diff
@@ -58,6 +59,76 @@ def _str(n: float) -> str:  # pragma: no cover
...
+    ret["size"] = [int(k) for k in ret["size"]]
+    return ret
+
+
+def _merge_and_sort_scales(
+    existing_scales: Sequence[dict[str, Any]], new_scales: Sequence[dict[str, Any]]
+) -> Sequence[dict[str, Any]]:
+    """Merge two list of scales, overwriting the old with new entries"""
+    ret_dict = dict(zip([k["key"] for k in existing_scales], existing_scales))
+    ret_dict.update(dict(zip([k["key"] for k in new_scales], new_scales)))
```
supersergiy, line 123: Could the two lines above could be replaced by slightly more readable

```python
ret_dict = {k['key']: k for k in existing_scales + new_scales}
```

supersergiy (changes requested):
Solid work! 

And yeah don't worry about the typing errors -- it is caused by a mypy update. We have a fix that's about to be merged into `main`, and which point it'll let us merge this branch in.

tests/unit/layer/volumetric/precomputed/test_precomputed.py
```diff
@@ -102,3 +104,225 @@ def test_ts_set_voxel_set_voxel_offset_chunk_and_data(clear_caches, mocker):
...
+                "voxel_offset": [4, 5, 6],
+            },
+        ],
+    )
+    info_spec.update_info(tmp_layer, on_info_exists="overwrite")
+    scales = get_info(tmp_layer)["scales"]
+    assert "asdf8" in [e["key"] for e in scales]
+    assert "33_33_40" in [e["key"] for e in scales]
+    for scale in scales:
+        if scale["key"] == "asdf8":
```
dodamih, line 204: Hmm, do we want to even allow random keys, or add a check? I'm not sure what the use case would be for arbitrary strings.

trivoldus28: I think `key` here is just a path for one particular scale? It could be useful for someone to have multiple scales of the same resolution, for example at different chunk sizes

zetta_utils/layer/volumetric/precomputed/precomputed.py
trivoldus28: Yea good call, as neuroglancer implements it the scale order does affect how it fetches the different resolutions

tests/unit/layer/volumetric/precomputed/test_precomputed.py
dodamih: Oh, I hadn't thought of that. Yeah, that sounds like something pretty useful - the only issue is that to support that we would need to add support for getting things like chunk size and offsets in a way that's decoupled from the resolution (right now, `get/set_offset`, `get/set_chunk_size` all assume X_Y_Z format for the key, and the rest of the codebase just goes off of VolumetricIndex.)

trivoldus28: @dodamih Good points but the workflow need not integrate every weird use cases. For general usage, having decoupled `key` could be useful when it's not easy to rename folders (e.g., prefixes in buckets) and the user somehow wanted to have multiple versions of the same resolution. This use case is pretty weird though and you're probably right we should put a check to make sure `key` follows X_Y_Z format until there is demand otherwise.

trivoldus28: @supersergiy @dodamih Do we even support selecting chunk sizes if there are options?

nkemnitz: >could be useful for someone to have multiple scales of the same resolution, for example at different chunk sizes

That edge case is covered somewhere else in the info file: The `scale['chunk_sizes']` expects a list of `[sx,sy,sz]` chunk sizes. I believe the intention was to let Neuroglancer pick the chunking which best fits the currently visible cross-section. But never tested it so far.

trivoldus28: Thanks! For now I think I'll restrict `key` to follow `x_y_z` res to ensure best compatibility with the rest of the framework.

trivoldus28:
@dodamih @supersergiy @nkemnitz One cool thing I came across is that you can actually have `..` in `key`! [[ng doc](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/volume.md#info-json-file-specification)].

trivoldus28:
@supersergiy besides that do you need me to do anything else to merge?

supersergiy (approved):
💯

PR #514, opened 2023-09-18
feat: preserve tuples in JSON

nkemnitz:
With this PR, tuples will be JSON encoded as list with a `__tuple__` marker as first element. The tuple preservation is required for libraries that differentiate between Python tuples and lists, i.e. `imgaug` treats tuples as value ranges to pick from, but JSON silently converts them to lists (discrete values)

Slows down JSON encoding/decoding a bit, but I don't think we use JSON for high-performance operations, anyway.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,7 +152,7 @@ max-line-length = 99
 max-locals = 25
 
 [tool.pylint.basic]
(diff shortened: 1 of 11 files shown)


supersergiy (approved):
Looks solid!

PR #515, opened 2023-09-18
feat(k8s): allow distinct resource limits/requests

nkemnitz:
Current behavior uses the worker_resources as both, lower and upper bound.
This PR allwos specifying a separate lower bound. (E.g. I want to ensure my process has 1 vCPU, but not throttle it if it goes beyond).

I tried to make it as backward compatible as possible, which is why the old name and argument order is preserved for public methods, but not sure if that causes confusion?


akhileshh (approved):
This looks like the best possible solution for now.

PR #526, opened 2023-10-18
pytorch-lightning upgrade

nkemnitz:
Had to disable multiprocessing for the jit/onnx export to work at all. I don't think it's a big slowdown - depends on the validation intervall, of course.

Upgrade is probably a good idea anyway, but main reason is the newer pytorch-lightning gave more reliable exceptions when testing DDP (and there may be a chance it fixes bugs with DDP we are unaware of).

Necessary changes in more detail:
* `validation_epoch_end` is now `on_validation_epoch_end` and has no return type
* `pytorch_lightning.utilities.distributed import distributed_available` was removed - it's was a one-liner I added to zetta_utils
* `pytorch_lightning.utilities.cloud_io` was moved to `lighning_fabric.utilities.cloud_io` (and got typings!)
* `pl.callbacks` is intialized later, which caused mypy to complain in our init. I moved the progress and checkpointing callbacks around to get initialized properly with the other PL stuff. `ConfigureTraceCallback` remains a special case, though :(
* Like mentioned above, had to disable the multiprocessing, otherwise the newly spawned process would attempt to login to wandb again and crash... but `fork` doesn't work with PyTorch
* some changes to how the TraceCallback collects the submodule names (`named_children()` rather than iterating through the attributes)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ segmentation = ["onnx >= 1.13.0", "onnxruntime-gpu >= 1.13.1"]
 training = [
     "zetta_utils[tensor_ops,cloudvol,convnet,viz,gcs,augmentations]",
     "torch >= 2.0",
-    "pytorch-lightning == 1.7.7",                       # mypy error on newer
+    "pytorch-lightning ~= 2.0.0",  # mp dataloader causes segmentation faults with 2.1.0
     "torchmetrics == 0.11.4",
     "wandb >= 0.13.1",
 ]
--- a/zetta_utils/training/lightning/regimes/alignment/base_encoder.py
+++ b/zetta_utils/training/lightning/regimes/alignment/base_encoder.py
@@ -41,7 +41,7 @@ class BaseEncoderRegime(pl.LightningModule):  # pylint: disable=too-many-ancesto
         optimizer = torch.optim.Adam(self.parameters(), lr=self.lr)
         return optimizer
 
-    def validation_epoch_end(self, _):
+    def on_validation_epoch_end(self):
         log_results(
             "val",
             "worst",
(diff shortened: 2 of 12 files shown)


nkemnitz:
I just checked the current main branch and noticed that it didn't export jit or onnx, either.

Think it's [this line](https://github.com/ZettaAI/zetta_utils/pull/526/files#diff-d7d24b371be4f087d00da084cea4983bbfe8e1d72f12ea6e737c6233d8cae702L232), which checked for `self.trainer.trace_configuration is None`, even though it's initialized and reset to `{}`? @dodamih , do you know what's going on here?

supersergiy (commented):
I think multiprocessing was added for JIT tracing because of a memory leak in pytorch. @dodamih could you confirm?

dodamih:
@nkemnitz Oh, that does look like a bug. And yeah, we did add multiprocessing for JIT tracing due to a memory leak - the only way we could do it was to insulate it as an outside process.

dodamih (approved):

supersergiy (changes requested):
Let's not merge until the leak is addressed

nkemnitz:
@supersergiy Took a while, but I think I got it now.
* `wandb` log operations are now only called from the PL `Trainers` - also modified the regimes accordingly
* source code is now also only logged once - by the rank=0 logger
* torch.jit and onnx export are now functional again (not working on `main`)
* added comment about the memory leak with consecutive jit.traces
* had to pin PL to 2.0.x - with 2.1.0 and multiple data loaders.I am getting segmentation faults after the first validation epoch

Successfully ran single node, multi-GPU training and 8 data loaders for multiple training epochs
Successfully ran multi node, single-GPU training with "0" data loaders. Multiple data loaders still won't work, but that's not because of the PL update.

"Successfully" here just means, it didn't crash and the graphs looked plausible... Some logging details might still be off, such as which workers are logging images.

supersergiy (approved):
💯

PR #533, opened 2023-10-25
[DO NOT MERGE] feat: Pipelining with workers

dodamih:


supersergiy:
Hm do you know why this has Tris commits?

PR #534, opened 2023-10-29
DDP fixes

nkemnitz:
* added @akhileshh's suggestion for the main node address.
* found a way to name the WandbLogger runs based on their global rank
* fixed a bug I introduced in the last PR: each DDP device must perform the exact same set of operations, otherwise you get `CollectiveFingerPrint` mismatches. More specifically: All devices must run the TraceCallback once, even if it's the exact same result and only rank 0 writes the checkpoint.


supersergiy (approved):
Thanks for the fixes!

PR #550, opened 2023-11-21
Add defect and resin detector

trivoldus28:


supersergiy (approved):
Let's merge this as is for now, but I'm wondering if the old-school tiling in the processor is actually any better than subchunkable here

PR #566, opened 2023-12-02
parallel `builder` mode + extra imports for `zetta run`

supersergiy:


supersergiy:
@nkemnitz any thoughts?

nkemnitz:
Tried it with [one of the misd training specs](https://github.com/ZettaAI/zetta_utils/blob/nkem/train-misd/specs/nico/training/aced_misd/train/64nm/1_5px/z1_enc_misd.cue), but got

`PicklingError: Can't pickle <function build_layer_set at 0x7fd4425f81f0>: it's not the same object as zetta_utils.layer.layer_set.build.build_layer_set`

supersergiy:
Thanks @nkemnitz. This looks interesting, will have to dig

nkemnitz (approved):
Tested with the encoder training spec - previous issue fixed and cut the time for loading layers in half on a 4 vCPU machine. (After actually enabling `parallel=True` in `train.py`)

nkemnitz (approved):

PR #568, opened 2023-12-06
feat(train): adds node_selectors based on provision_model

akhileshh:
Also moved creation of `dshm` and `tmp` volumes because they were repeated, and pylint was complaining about having too many statements in `_lightning_train_remote`

### DIFF
--- a/zetta_utils/training/lightning/train.py
+++ b/zetta_utils/training/lightning/train.py
@@ -239,7 +239,10 @@ def _get_tolerations() -> List[k8s_client.V1Toleration]:
     gpu = k8s_client.V1Toleration(
         key="nvidia.com/gpu", operator="Equal", value="present", effect="NoSchedule"
     )
-    return [gpu]
+    worker = k8s_client.V1Toleration(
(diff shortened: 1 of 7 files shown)


zetta_utils/training/lightning/train.py
```diff
@@ -364,11 +354,8 @@ def _lightning_train_remote(
         restart_policy="Never",
         # with multinode ddp we need a node on standard pool for an IP
         # that remains the same for the duration of training
-        # these pools must have `master-pool=true` taint
-        # not ncessary for single node ddp so it can be scheduled on preemptibles
-        tolerations=_get_tolerations(
-            "worker"
-        ),  # _get_tolerations(role="master" if num_nodes > 1 else "worker"),
+        node_selector={"cloud.google.com/gke-provisioning": "standard"},
```
supersergiy, line 357: this should switch to the user selected prov model when num_nodes == 1, right?

akhileshh: done

PR #570, opened 2023-12-06
adds callback to register wandb logger

### LINKED ISSUE #559: New remote training causes double wandb initialization
The training spec is first build (and wandb initialized) on scheduler side, as well as on the remote node(s).

Some of the specs (e.g. encoder training) are already slow to build due to all the different annotaton layers. Doing so locally doubles the time before training starts.

The wandb initialization on scheduler side causes garbled wandb system logs (because both keep logging): 
image

akhileshh:
fixes #559 and #561


supersergiy (approved):
💯

PR #584, opened 2023-12-13
kisuk_playground: Data Pipeline Changes

supersergiy:


zetta_utils/layer/volumetric/tools.py
```diff
@@ -121,6 +133,53 @@ def process_data(self, data: torch.Tensor, mode: Literal["read", "write"]) -> to
         return result
 
 
+@builder.register("ROIMaskProcessor")
+@typechecked
+@attrs.mutable
+class ROIMaskProcessor(JointIndexDataProcessor):  # pragma: no cover
```
dodamih, line 139: This needs a comment since the `_mask` naming convention is hardcoded.

dodamih (changes requested):

zetta_utils/layer/volumetric/tools.py
supersergiy: Good catch. @torms3 could you add a comment explaining how to use this guy?
I'll write a test for `process_data` in the meanwhile, it's too complicated for `no cover`

supersergiy: Done with the test: https://github.com/ZettaAI/zetta_utils/pull/584/commits/97fe120e113a66b3784d6ada5f8a84525f388d31

torms3: I added a docstring: [``2ed8dad``](https://github.com/ZettaAI/zetta_utils/pull/584/commits/2ed8dad5a6e0cae36cba9e406ce0097d4fcf399f)

How does it sound? @dodamih @supersergiy

supersergiy: Thanks, looks great!

PR #593, opened 2023-12-28
fix(id_generation): recursive hashing of class methods and custom attribs

nkemnitz:
Should now work for most custom classes, not just `TaskableOperation` and `partial` / `BuilderPartial`. Performance might be worse, though.


trivoldus28:
Nice I'll give it a test

PR #598, opened 2024-01-03
Update GKE command

nkemnitz:
* replaces the default service account using `--scopes` with our workload identity service account. (Caused issues with DDP at some point, I believe)
* remove duplicate `--num-nodes` entries
* use `latest` GPU driver daemonset by default. Required for using L4 and other newer GPU, and recently also a Pytorch 2.1 issue @torms3 ran into.


supersergiy (approved):
Thanks for the fix!

PR #600, opened 2024-01-03
feat(train): start remote GPU training from instance without GPU

nkemnitz:
The CPU scheduler node might still fail to load the weights file during initialization, e.g. if it contains unsupported data types. This may be a scenario where we just want to skip the sanity checks on scheduler side entirely...

### DIFF
--- a/zetta_utils/convnet/utils.py
+++ b/zetta_utils/convnet/utils.py
@@ -90,7 +90,9 @@ def load_weights_file(
         return model
 
     with fsspec.open(ckpt_path) as f:
-        loaded_state_raw = torch.load(f)["state_dict"]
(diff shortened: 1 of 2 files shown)


supersergiy (approved):
Great, thanks for the fix! Im not sure if understand the description - does skipping sanity checks have any advantages over what you've implemented here?

supersergiy:
Is it about layers without CPU implementations?

nkemnitz:
The CPU-only scheduler complained when loading a model that was trained with fp16 that that is not supported on CPU. That was just a warning, though. I don't know if there are other situations where the CPU-only scheduler outright refuses to start remote(!) training because of such a falled check. That would be dumb, because I absolutely don't care what the scheduler thinks about my model if remote training is done on GPU.

PR #604, opened 2024-01-05
add weight_map to get_rigidity_map_zcxy function

JoeStrout:
...and a benchmark script to measure its performance


supersergiy (approved):
💯

PR #605, opened 2024-01-10
Joe/quickstart tweaks

JoeStrout:
- fixed typo in examples doc
- added reminder about installing `cue` in quick_start
- added better runtime error message when `cue` is not installed
- added level_intermediaries_dirs to initial example to make it run without error


supersergiy (approved):
💯

PR #610, opened 2024-01-16
Deconflict intermediary folder names by adding a random identifier

### LINKED ISSUE #551: Subchunkable: Non-checkerboarded intermediaries result in a FileNotFoundError on `delete_if_local`.
Happens probabilistically, whether local or remote, single process or multi-process.

dodamih:
Closes #551.


supersergiy (commented):
Should a unique ID be made for each subchunkable invocation, or should the children inheret it from the parents?

Also, should we use MUID instead of UUID? Easier if we ever have to look at it for debug

dodamih:
It should be inherited, which is what it's doing - each level has its own folder, so there shouldn't be a conflict.

I can switch to MUID.

supersergiy (approved):
Got it, thanks for explaining. Feel free to either use MUID or not

PR #611, opened 2024-01-16
feat: misalignment augmentation

supersergiy:


zetta_utils/augmentations/misalign.py
```diff
@@ -0,0 +1,162 @@
...
+            start_offset[1] = -disp_y_in_idx_res
+        else:
+            end_offset[1] = -disp_y_in_idx_res
+
+        idx = idx.translated_start(Vec3D[int](*start_offset)).translated_end(
+            Vec3D[int](*end_offset)
+        )
+        self.prepared_disp_fraction = (
+            disp_x_in_idx_res / idx.shape[0],
+            disp_y_in_idx_res / idx.shape[0],
```
torms3, line 80: Shouldn't it be `idx.shape[1]`?

zetta_utils/augmentations/misalign.py
```diff
@@ -0,0 +1,162 @@
...
+        if isinstance(data, torch.Tensor):
+            z_size = data.shape[-1]
+        else:
+            if not self.keys_to_apply:
+                raise ValueError(
+                    "`keys_to_apply` must be a non-empty list of strings when "
+                    "applying to data of type `dict`"
+                )
+            sizes = [data[k].shape for k in self.keys_to_apply]
+            if not all(e[-1] == sizes[0][-1] for e in sizes):
```
torms3, line 97: I typically use an output patch size smaller than the input patch size, so we should allow varying z sizes.

zetta_utils/augmentations/misalign.py
```diff
@@ -0,0 +1,162 @@
...
+        return idx
+
+    def process_data(self, data: T, mode: Literal["read", "write"]) -> T:
+        if mode != "read":
+            raise NotImplementedError()
+
+        if isinstance(data, torch.Tensor):
+            z_size = data.shape[-1]
+        else:
+            if not self.keys_to_apply:
```
torms3, line 91: Misalignment is a "global" augmentation, meaning it should be applied to all input/output volumes in a typical use case. Requiring users to specify the list of keys can be error-prone. Therefore, I think it's better to default to applying it to all keys.

supersergiy: Step misalignment is global, but slip should only be seen through so it doesn't affect all layers. Making default behaviors differ for slip and step is misleading. This behavior was chosen because explicit is generally preferable to implicit.

zetta_utils/augmentations/misalign.py
```diff
@@ -0,0 +1,165 @@
...
+            z_size = max(size[-1] for size in sizes)
+
+        z_chosen = random.randint(0, z_size - 1)
+
+        def _process_tensor(tensor: torch.Tensor) -> torch.Tensor:
+            if tensor.size(-1) < z_size:
+                z_diff = z_size - tensor.size(-1)
+                assert z_diff > 0
+                assert z_diff % 2 == 0
+                this_z_chosen = z_chosen - (z_diff // 2)
```
torms3, line 114: The logic here is broken. If `z_chosen` is 0, then `this_z_chosen` becomes negative. Creating a `slice` with a negative `z_chosen` is a logical error.

zetta_utils/augmentations/misalign.py
torms3: The "see through" mode for the slip misalignment is not the default; it's just an option. 

I would still argue that requiring users to specify the list of keys can be highly inconvenient and error-prone. The key point is as follows: it is highly preferable to make data augmentation modular and independent of other components of the experiment. For instance, it is very common in practice to add or remove target layers (labels) from experiment to experiment, especially when it comes to multi-task learning, e.g., embeddings, affinities, mitochondria, semantic segmentation, myelin, etc. Requiring users to specify the list of keys creates an interdependency between the data augmentation and other components, so whenever you experiment with different multi-task learning setups, you have to manually modify the list of keys every time. This is the situation that I would most like to avoid in practice.

I think the cleanest solution here would be to make `keys_to_apply` optional.

zetta_utils/augmentations/misalign.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def process_index(
+        self, idx: VolumetricIndex, mode: Literal["read", "write"]
+    ) -> VolumetricIndex:
+        if mode != "read":
+            raise NotImplementedError()
+        disp_x_in_unit_magn = random.uniform(self.disp_min_in_unit, self.disp_max_in_unit)
+        disp_y_in_unit_magn = random.uniform(self.disp_min_in_unit, self.disp_max_in_unit)
+        if self.disp_in_unit_must_be_divisible_by is not None:
+            disp_x_in_unit_magn = math.floor(
```
torms3, line 50: I'm totally confused here. If the purpose of `disp_in_unit_must_be_divisible_by` is as the name of the variable implies, then shouldn't it be
```
disp_x_in_unit_magn = (
        disp_x_in_unit_magn // self.disp_in_unit_must_be_divisible_by
    ) * self.disp_in_unit_must_be_divisible_by
```
?

zetta_utils/augmentations/misalign.py
supersergiy: All of the target layers can be defined in a list in a CUE file as a variable, and used in both places so that the user doesn't have to duplicate everything. Making `keys_to_apply` optional will make for a scenario where a slip missalignment will be applied to all layers. They user may not realize it, and the network will be trained with a bad setting, which could be an extremely difficult error to notice after the fact. I think explicit-over-implicit is cleaner here -- the user should specify what layer they want to shift explicitly, as it is important.

supersergiy: The alternative is to make behaviors differ for slip vs step. I think it is acceptable. As long as the explicit keys are required for slip

zetta_utils/augmentations/misalign.py
```diff
@@ -0,0 +1,217 @@
...
+            disp_y_in_unit_magn = (
+                disp_y_in_unit_magn // self.disp_in_unit_must_be_divisible_by
+            ) * self.disp_in_unit_must_be_divisible_by
+        disp_x_in_unit = disp_x_in_unit_magn * _choose_pos_or_neg_misalignent()
+        disp_y_in_unit = disp_y_in_unit_magn * _choose_pos_or_neg_misalignent()
+
+        disp_x_in_idx_res = disp_x_in_unit / idx.resolution[0]
+        disp_y_in_idx_res = disp_y_in_unit / idx.resolution[1]
+
+        # self.prepared_disp = (disp_x, disp_y)
```
torms3, line 76: Why was this commented-out line left?

zetta_utils/augmentations/misalign.py
```diff
@@ -0,0 +1,217 @@
...
+        keys_to_apply_final: list[str] | None = None
+        if isinstance(data, torch.Tensor):
+            z_size = data.shape[-1]
+        else:
+            keys_to_apply_final = self._get_keys_to_apply_final(data)
+            sizes = [data[k].shape for k in keys_to_apply_final]
+            if self.disp_in_unit_must_be_divisible_by is None and not all(
+                e[0:2] == sizes[0][0:2] for e in sizes
+            ):
+                raise ValueError(
```
torms3, line 201: I don't fully understand this constraint. The composition of misalignment augmentation is widely used, and misalignment augmentation can always make the volumetric index non-square.

Issue #452, opened 2023-07-30
Floating point data resolutions causing issues.


nkemnitz:
Ran into a bunch of issues related to dealing with datasets that use non-integer voxel resolutions.

The usual suspects are
* flooring of floating point values representing integers after division
* using exact comparison between (differently) calculated floating point values

I fixed the ones I encountered on Friday in #451, but I wonder if we can maybe find a better approach?

Toughts so far:
* Use `decimal`
* Just scale resolutions by a factor of 10, 100 or 1000 and prohibit floating point resolutions (and bounds) completely within zutils...

supersergiy:
@dodamih and I have discussed using fixed point back in the day.
@dodamih any thoughts on this now?

dodamih:
If this is causing issues, we should switch to fixed point. However, one issue with `decimal` is that the precision seems to be set from the context, which I'm not sure would work well for inferring return types.

One other option is to roll our own similar to what Nico's suggesting that behaves something like `decimal` - e.g. `float` is allowed, but Vec3D will always return something rounded to X digits below decimal. The benefit of this is that this is something that's internal to Vec3D. Thoughts?

supersergiy:
From the meeting:

@nkemnitz doesn't like Decimal -- no other library expects decimals, and nor do our users.
@dodamih suggested a solution where Vec3D has `precision` attribute, and arithmetic operations will be rounded when one of the operands has precision. @supersergiy is worried about users pulling out individual components of Vec3D and doing lossy operations on top of them.

If we do go with some fixed-point approach (Decimal or integer that we interpret as a fixed-point), we think that only stuff measured in nanometers needs to be fixed point. So that means bounding cubes and resolutions... is there anything else, or just these two?

The operations that I know of that we do with these two are:
1. Divide bcube during chunking
2. Multiply resolution by a `res_change_mult`
How exhaustive is this list? Do you know of other operations that would need to be protected?

supersergiy:
Is this confirmed to be a non-issue anymore? cc @nkemnitz

nkemnitz:
There are a few more places that ignore rounding issues, e.g. https://github.com/ZettaAI/zetta_utils/blob/c48ac5e3837f528e03159be4ee0f9d3d3902549c/zetta_utils/training/datasets/sample_indexers/volumetric_ngl_indexer.py#L67

supersergiy:
@nkemnitz can this be closed?

nkemnitz:
[...]

PR #613, opened 2024-01-16
fix(ngl_indexer): floating point resolution rounding

nkemnitz:
Resolves another issue mentioned in #452.


supersergiy (approved):
looks good, thanks Nico

PR #615, opened 2024-01-17
hotfix: build list/dict of BuilderPartial objects

nkemnitz:
* fixes `zu.builder.build([{"@type": "to_uint8", "@mode": "partial"}]` returning an empty list
* fixes `dill.dumps(zu.builder.build([{"@type": "to_uint8", "@mode": "partial"}]), recurse=True` failing with `RuntimeError: _SafeQueue objects should only be shared between processes through inheritance`
----

* **not** fixed: `zetta run -p` will still fail for complex flows with e.g. `PicklingError: Can't pickle <function concurrent_flow at 0x7fe514f87f70>: it's not the same object as zetta_utils.mazepa.flows.concurrent_flow`

### DIFF
--- a/zetta_utils/builder/build.py
+++ b/zetta_utils/builder/build.py
@@ -6,6 +6,7 @@ from concurrent.futures import ProcessPoolExecutor
 from typing import Any, Callable, Final, Optional, cast
 
 import attrs
+import cachetools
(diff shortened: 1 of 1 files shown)


supersergiy (approved):
👍

PR #617, opened 2024-01-19
Joe/optional dst

JoeStrout:
Adds support for specifying `None` as the `dst` parameter to build_subchunkable_apply_flow.  Doing so also requires skip_intermediaries = True, and expand_bbox_backend = False.

A test CUE file is at https://gist.github.com/JoeStrout/14879c2c923ec63925243559d661f47d, but I will need some help making a test that models a more realistic use-case (this one is just a dummy copy op).

### DIFF
--- a/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
+++ b/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
@@ -79,7 +79,7 @@ class DelegatedSubchunkedOperation(Generic[P]):
 
 @builder.register("build_subchunkable_apply_flow")
 def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg, too-many-locals, too-many-branches, too-many-statements
-    dst: VolumetricBasedLayerProtocol,
+    dst: VolumetricBasedLayerProtocol | None,
     dst_resolution: Sequence[float],
     processing_chunk_sizes: Sequence[Sequence[int]],
     processing_crop_pads: Sequence[int] | Sequence[Sequence[int]] = (0, 0, 0),
@@ -114,8 +114,10 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
     blending. Chunks are processed, written to an intermediary location temporarily, and then
     combined (``reduced``) with weights if necessary to produce the output.
 
-    :param dst: The destination VolumetricBasedLayerProtocol.
-    :param dst_resolution: The resolution of the destination VolumetricBasedLayerProtocol.
+    :param dst: The destination VolumetricBasedLayerProtocol.  May be None, in which case
+        `skip_intermediaries` must be True, and `expand_bbox_backend` must be False.
+    :param dst_resolution: The resolution of the destination VolumetricBasedLayerProtocol
+        (or resolution to use for computation, even if `dst` is None).
     :param processing_chunk_sizes: The base chunk size at each subchunking level in X, Y, Z,
         from the largest to the smallest. Subject to divisibility requirements (see bottom). When
         ``auto_divisibility`` is used, the chunk sizes other than the bottom level chunk size will
@@ -209,6 +211,12 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
                 " `coord_resolution` cannot be specified."
             )
         bbox_ = bbox
+
+    if dst is None:
+        if not skip_intermediaries:
+            raise ValueError("`skip_intermediaries` must be True when `dst` is None.")
+        if expand_bbox_backend:
+            raise ValueError("Cannot use `expand_bbox_backend` when `dst` is None.")
 
     if fn is not None and op is not None:
         raise ValueError("Cannot take both `fn` and `op`; please choose one or the other.")
@@ -545,7 +553,7 @@ def _make_ng_link(
 
 
 def _print_summary(  # pylint: disable=line-too-long, too-many-locals, too-many-statements
-    dst: VolumetricBasedLayerProtocol,
+    dst: VolumetricBasedLayerProtocol | None,
     dst_resolution: Vec3D,
     level_intermediaries_dirs: Sequence[str | None],
     skip_intermediaries: bool,
@@ -592,10 +600,14 @@ def _print_summary(  # pylint: disable=line-too-long, too-many-locals, too-many-
         summary += lrpad(f"Volume of ROI: {vol*1e-18:10.3f} mm^3", 1, length=120) + "\n"
     summary += lrpad(length=120) + "\n"
     summary += lrpad("Output location(s):", 1, length=120) + "\n"
-    layer_strs = dst.pformat().split("\n")
+    if dst is None:
+        summary += lrpad("`dst` is None\n")
+        layer_strs = []
+    else:
+        layer_strs = dst.pformat().split("\n")
     for layer_str in layer_strs:
         summary += lrpad(layer_str, 2, length=120) + "\n"
-    if generate_ng_link:
+    if generate_ng_link and dst is not None:
         ng_link = _make_ng_link(dst, bbox)
         if ng_link is not None:
             summary += lrpad(length=120) + "\n"
@@ -674,7 +686,7 @@ def _print_summary(  # pylint: disable=line-too-long, too-many-locals, too-many-
 
 
 def _build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg, line-too-long, too-many-locals, too-many-branches, too-many-statements
-    dst: VolumetricBasedLayerProtocol,
+    dst: VolumetricBasedLayerProtocol | None,
     dst_resolution: Vec3D,
     level_intermediaries_dirs: Sequence[str | None],
     skip_intermediaries: bool,
@@ -706,7 +718,7 @@ def _build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg
     if expand_bbox_resolution:
         bbox = _expand_bbox_resolution(bbox, dst_resolution)
 
-    if expand_bbox_backend:
+    if expand_bbox_backend and dst is not None:
         bbox = _expand_bbox_backend(bbox, dst, dst_resolution)
 
     if shrink_processing_chunk:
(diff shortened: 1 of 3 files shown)


zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -114,8 +114,10 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
     blending. Chunks are processed, written to an intermediary location temporarily, and then
     combined (``reduced``) with weights if necessary to produce the output.
 
-    :param dst: The destination VolumetricBasedLayerProtocol.
-    :param dst_resolution: The resolution of the destination VolumetricBasedLayerProtocol.
+    :param dst: The destination VolumetricBasedLayerProtocol.  May be None, in which case
+        skip_intermediaries should be true, and expand_bbox_backend should be false.
```
supersergiy, line 118: it might be more correct to say "must" instead of "should"

zetta_utils/mazepa_layer_processing/common/volumetric_callable_operation.py
```diff
@@ -94,6 +94,12 @@ def __call__(  # pylint: disable=keyword-arg-before-vararg
                     semaphore_stack.enter_context(semaphore(semaphore_type))
             result_raw = self.fn(**task_kwargs)
             torch.cuda.empty_cache()
+
+        # If no destination layer, we can bail out now.  Possibly we
+        # could bail out a bit sooner.  ToDo: study this more.
+        if dst is None:
+            return
```
supersergiy, line 101: could you overwrite it as `if dst is not None: ... # rest of the code`
as a rule of thumb we try to minimize the number of return statements in our functions

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -40,10 +40,12 @@ def __call__(
         dst: VolumetricBasedLayerProtocol,
         idx: VolumetricIndex,
     ) -> None:
-        with semaphore("read"):
-            data = src[idx]
-        with semaphore("write"):
-            dst[idx] = data
+        if src is not None:
```
supersergiy, line 43: I think neither src nor dst can be None here. This op expect both src and dst, otherwise it's a user error

JoeStrout: OK, thanks for the review!

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -79,7 +79,7 @@ def __call__(
 
 @builder.register("build_subchunkable_apply_flow")
 def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg, too-many-locals, too-many-branches, too-many-statements
-    dst: VolumetricBasedLayerProtocol,
+    dst: Optional[VolumetricBasedLayerProtocol],
```
dodamih, line 82: I suggested incorrectly before, | `None` is cleaner here.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -210,6 +212,12 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
             )
         bbox_ = bbox
 
+    if dst is None:
+        if not skip_intermediaries:
+            raise ValueError("skip_intermediaries must be True when dst is None.")
```
dodamih, line 217: "`skip_intermediaries` must be True when `dst` is None." would be better here. (Just backticks for variable name.)

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -210,6 +212,12 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
...
+        if expand_bbox_backend:
+            raise ValueError("Cannot use expand_bbox_backend when dst is None.")
```
dodamih, line 219: Same thing with variable name in backticks for `expand_bbox_backend`.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -810,18 +822,21 @@ def _build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg
...
-    if skip_intermediaries is True:
-        logger.info(
-            "Since intermediaries are skipped, the ROI and any writing done to the final destination are "
-            "required to be chunk-aligned."
-        )
-        dst = deepcopy(dst).with_changes(
-            backend=dst.backend.with_changes(enforce_chunk_aligned_writes=True)
-        )
+    if dst is None:
+        logger.info("dst is None; skipping generation of standard output")
```
dodamih, line 826: `dst` in backticks, and period at the end.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -434,6 +434,7 @@ def make_tasks_with_checkerboarding(  # pylint: disable=too-many-locals, too-man
             backend chunk size to half of the processing chunk size. Furthermore, skip caching
             if the temporary destination happens to be local.
             """
+            assert dst is not None  # (impossible to get a temp dst without dst and intermediaries)
```
dodamih, line 437: This doesn't need a comment - if we see `assert foo is not None` in the code, we know that it's to get around typing for an optional kwarg. (The context is that we want to keep the number of comments in the code to a minimum.)

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -545,6 +546,7 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
             logger.info(f"Submitting {len(tasks)} processing tasks from operation {self.op}.")
             yield tasks
         elif not self.use_checkerboarding and self.force_intermediaries:
+            assert dst is not None  # no intermediaries without destination layer
```
dodamih, line 549: Also unnecessary.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -569,10 +573,9 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
             red_chunks = reduction_chunker(
                 idx, mode="exact", stride_start_offset_in_unit=stride_start_offset_in_unit
             )
+            reduction_dst = None if dst is None else dst.with_procs(read_procs=(), write_procs=())
             tasks_reduce = [
-                Copy().make_task(
-                    src=dst_temp, dst=dst.with_procs(read_procs=(), write_procs=()), idx=red_chunk
-                )
+                Copy().make_task(src=dst_temp, dst=reduction_dst, idx=red_chunk)
```
dodamih, line 578: This is weird - is any of this change necessary? We've already asserted that `dst is not None` above.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -585,6 +588,9 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
             delete_if_local(dst_temp)
         # case with checkerboarding
         else:
+            assert (
+                dst is not None
+            )  # since checkerboarding requires intermediaries, which requirest dst
```
dodamih, line 593: Comment unnecessary.

dodamih (changes requested):

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
JoeStrout: I suspect that when we (or some new guy) come back to this code 3 years from now, we will no longer remember why it is safe to assert that dst is not None here (or in the previous case).  But I trust your judgement; out it goes.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
JoeStrout: Yeah this was left over from earlier work, and I didn't catch it.  You are right.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
dodamih: That's a good point, but I think this needs to actually be caught in `subchunkable_apply_flow.py` - if `use_checkerboarding` (which is inferred) or `force_intermediaries` (given directly) is set when `dst` is None, that should raise an exception before it gets to this stage.

dodamih: (Theoretically, the user may use `VolumetricApplyFlow` as a standalone, but that's not really expected - we used to have a `build_` method in there but removed it with the expectation that users will use subchunkable instead.)

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -114,8 +114,10 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
     blending. Chunks are processed, written to an intermediary location temporarily, and then
     combined (``reduced``) with weights if necessary to produce the output.
 
-    :param dst: The destination VolumetricBasedLayerProtocol.
-    :param dst_resolution: The resolution of the destination VolumetricBasedLayerProtocol.
+    :param dst: The destination VolumetricBasedLayerProtocol.  May be None, in which case
+        skip_intermediaries must be True, and expand_bbox_backend must be False.
```
dodamih, line 118: skip_intermediaries in backticks.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -592,10 +600,14 @@ def _print_summary(  # pylint: disable=line-too-long, too-many-locals, too-many-
         summary += lrpad(f"Volume of ROI: {vol*1e-18:10.3f} mm^3", 1, length=120) + "\n"
     summary += lrpad(length=120) + "\n"
     summary += lrpad("Output location(s):", 1, length=120) + "\n"
-    layer_strs = dst.pformat().split("\n")
+    if dst is None:
+        summary += lrpad("dst is None\n")
```
dodamih, line 604: dst in backticks

dodamih (approved):

PR #618, opened 2024-01-19
refactor: execution to run

akhileshh:


zetta_utils/cli/main.py
```diff
@@ -76,12 +83,20 @@ def validate_py_path(ctx, param, value):  # pylint: disable=unused-argument
     callback=validate_py_path,
     help="Specify additional imports. Must end with `.py`.",
 )
+@click.option(
+    "--no_heartbeat",
```
supersergiy, line 87: Not super sure but it might generate a `--no_heartbeat/--no_no_heartbeat`

zetta_utils/cli/main.py
```diff
@@ -103,10 +118,13 @@ def run(
     if parallel_builder:
         zetta_utils.builder.PARALLEL_BUILD_ALLOWED = True
 
-    result = zetta_utils.builder.build(spec, parallel=parallel_builder)
-    logger.debug(f"Outcome: {pprint.pformat(result, indent=4)}")
-    if pdb:
-        breakpoint()  # pylint: disable=forgotten-debug-statement # pragma: no cover
+    if no_heartbeat:
+        _heartbeat_interval = -1
```
supersergiy, line 122: Maybe `None` is cleaner?

supersergiy (commented):
Looks solid so far!

zetta_utils/cli/main.py
akhileshh: That is only when a `/` is added by the user - [boolean-flags](https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags).
Should we do a `@click.option('--heartbeat/--no-heartbeat', default=True)` then? Feels more intuitive.

supersergiy (approved):
Looks excellent!

PR #622, opened 2024-01-24
Fix kornia dilation from adding 1s

trivoldus28:
Current dilation behavior adds `1.0`/`True`'s around the border of the tensor which is unexpected:
```
>>> a
tensor([[[[False, False, False, False, False, False],
          [False, False, False, False, False, False],
          [False, False, False, False, False, False],
          [False, False, False, False, False, False],
          [False, False, False, False, False, False],
          [False, False, False, False, False, False]]]])
>>> dilation(a, kernel)
tensor([[[[1., 1., 1., 1., 1., 1.],
          [1., 0., 0., 0., 0., 1.],
          [1., 0., 0., 0., 0., 1.],
          [1., 0., 0., 0., 0., 1.],
          [1., 0., 0., 0., 0., 1.],
          [1., 1., 1., 1., 1., 1.]]]])
```

This patch sets the default border value to be `0.0`:
```
>>> dilation(a, kernel, border_type="constant")
tensor([[[[0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0.]]]])
```


supersergiy (approved):
Thanks for the fix!

PR #623, opened 2024-01-30
fix(testing_integration): resolve LFS pointers

nkemnitz:
The `orjson.JSONDecodeError` occured because the LFS pointers were not converted - so zutils attempted to read the pointers rather than the actual file content.
The integration tests still fail, but the first subchunkable tests are now passing again: https://github.com/ZettaAI/zetta_utils/actions/runs/7709264157/job/21010085787


supersergiy (approved):
Thanks for the fix

PR #624, opened 2024-01-30
feat(tools): Expand StartOffsetOverrider to IndexOverrider

nkemnitz:
Should be fully backward compatible (just have to rename `VolumetricStartOffsetOverrider` with `VolumetricIndexOverrider`). I haven't seen it being used in any cue file, though.


supersergiy (approved):
Looks great, thanks Nico

PR #626, opened 2024-01-30
fix(ReduceByWeightedSum): rounding for integer layers

nkemnitz:
Should resolve the failing integration test `uint8_copy_writeproc_multilevel_checkerboard` and weird Mandala artifacts in blended uint8 layers.

The backend layer data type was used causing truncation errors with every additional blended layer if the destination is using integer


nkemnitz:
mypy tests failing due to #627

nkemnitz:
Integration tests now all passing: https://github.com/ZettaAI/zetta_utils/actions/runs/7718099502

dodamih (approved):
Thank you so, so much!

PR #632, opened 2024-02-02
fix(subchunkable_apply_flow): generate deterministic flow_id

nkemnitz:
* made `generate_invocation_id` a bit more lenient regarding input parameter types... it doesn't really care about what objects go in, they all get serialized as a tuple
* added dedicated subchunkable flow test (with level intermediaries) to test_id_generation - it's one of the main objects used
* made the suggested change to identify the subchunkable_apply_flow based on its local variables


supersergiy (approved):
thanks for the fix Nico?

PR #634, opened 2024-02-02
feat: spawn pool once per execution

supersergiy:


trivoldus28:
Thanks for the fix! That should resolve the immediate issue. I'm wondering one thing though is whether `acknowledge_fn` is ever expected to fail or throw an exception? I'm just afraid that let's say if an implementation time outs or fails in anyway for any ack, it would bring down the process and the scheduler even though the error itself is benign and doesn't affect scheduling. In this case, it should be more fault tolerant to just spawn a thread for each `acknowledge_fn` invokation, and if it fails just prints the stack on stdout to let the user know and let the scheduler carry on.

supersergiy:
@trivoldus28 acknowledge_fn already has retry with random backoff. if it fails it should be an exception

trivoldus28 (approved):
lgtm, thanks!

PR #635, opened 2024-02-05
feat(gc): make resource centric

akhileshh:


zetta_utils/run/gc.py
```diff
@@ -25,24 +26,32 @@
...
 def _get_stale_run_ids() -> list[str]:  # pragma: no cover
-    client = RUN_DB.backend.client  # type: ignore
-
     lookback = int(os.environ["EXECUTION_HEARTBEAT_LOOKBACK"])
     time_diff = datetime.utcnow().timestamp() - lookback
 
-    filters = [("heartbeat", "<", time_diff)]
-    query = client.query(kind="Column", filters=filters)
-    query.keys_only()
+    resourcedb_client = RESOURCE_DB.backend.client  # type: ignore
```
supersergiy, line 42: Is there something we can add to DBLayer interface to avoid doing it through the backend here?

akhileshh: Add a `client` property - most (all?) databases have one.

supersergiy: there wouldn't be a single type of that property that we can specify.
we could make it `Any`, but then different backend types stop being interoperable.
what about adding a way to achieve the functionality that you're doing here through the DBLayer itself instead of the Backend?

akhileshh: For this specific use case it is possible, it is asking for `run_id` property from all rows. Similar to `select run_id from run_table;`, with no `where` clause.

We could use special character to mean read all rows/entities.

supersergiy (approved):
Looks good!

PR #637, opened 2024-02-06
fix: `local_test` without GCP ctx mngr

### LINKED ISSUE #630: `mazepa.execute_on_gcp_with_sqs` with `local_test: true` spawns k8s deployment
That must have happened in the last few weeks - `local_test: true` should definitely not create a k8s deployment.

supersergiy:
fixes #630


dodamih (approved):
Thanks for fixing this!

PR #639, opened 2024-02-07
feat: gap

JoeStrout:
Certainly incomplete, but at a point where I need some review/feedback from Dodam.

### DIFF
--- a/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
+++ b/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
@@ -107,6 +107,7 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
     start_coord: Sequence[int] | None = None,
     end_coord: Sequence[int] | None = None,
     coord_resolution: Sequence | None = None,
+    gap: Sequence[int] = (0, 0, 0),
 ) -> mazepa.Flow:
     """
     The helper constructor for a flow that applies any function or operation with a `Tensor`
@@ -194,6 +195,7 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
         and ``coord_resolution``; cannot be used with ``bbox``.
     :param coord_resolution: The resolution in which the coordinates are given for the bounding
         box. Must be used with ``start_coord`` and ``end_coord``; cannot be used with ``bbox``.
+    :param gap: Extra unprocessed space to be skipped between chunks at the top level.
     """
     if bbox is None:
         if start_coord is None or end_coord is None or coord_resolution is None:
@@ -330,6 +332,18 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
     else:
         allow_cache_up_to_level_ = allow_cache_up_to_level
 
+    if gap is not None and Vec3D(*gap) != Vec3D[int](0, 0, 0):
+        if any(v % 2 != 0 for v in gap):
+            raise ValueError("`gap` must be divisible by 2 in all dimensions.")
+        if any(v != 0 for v in processing_blend_pads_[0]):
+            raise ValueError("`blend_pads` at top level must be zero when `gap` is used.")
+        if shrink_processing_chunk:
+            raise ValueError("`shrink_processing_chunk` cannot be used with nonzero `gap`.")
+        if expand_bbox_backend:
+            raise ValueError("`expand_bbox_backend` cannot be used with nonzero `gap`.")
+        if expand_bbox_resolution:
+            raise ValueError("`expand_bbox_resolution` cannot be used with nonzero `gap`.")
+
     assert len(list(op_args)) == 0
     return _build_subchunkable_apply_flow(
         dst=dst,
@@ -353,6 +367,7 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
         generate_ng_link=generate_ng_link,
         op_args=op_args,
         op_kwargs=op_kwargs_,
+        gap=Vec3D(*gap),
     )
 
 
@@ -462,6 +477,7 @@ def _expand_bbox_processing(  # pylint: disable=line-too-long
     bbox: BBox3D,
     dst_resolution: Vec3D,
     processing_chunk_sizes: Sequence[Vec3D[int]],
+    gap: Vec3D[int],
 ) -> BBox3D:
     bbox_shape_in_res = round(bbox.shape / dst_resolution)
     bbox_shape_in_res_raw = bbox.shape / dst_resolution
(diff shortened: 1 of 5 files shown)


zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -330,6 +332,12 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
     else:
         allow_cache_up_to_level_ = allow_cache_up_to_level
 
+    if gap is not None and Vec3D(*gap) != Vec3D[int](0, 0, 0):
+        if any(v != 0 for v in processing_blend_pads_[0]):
+            raise ValueError("`blend_pads` at top level must be zero when `gap` is used.")
+        if shrink_processing_chunk:
+            raise ValueError("`shrink_processing_chunk` cannot be used with nonzero `gap`.")
+
```
dodamih, line 340: Missing check that `gap` is evenly divisible by 2 in all dimensions.

Also check for `expand_bbox_backend` / `expand_bbox_resolution` and raise as in the case of `shrink_processing_chunk`.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -391,6 +400,7 @@ def _auto_divisibility(  # pylint: disable=line-too-long
     processing_chunk_sizes: Sequence[Vec3D[int]],
     processing_crop_pads: Sequence[Vec3D[int]],
     processing_blend_pads: Sequence[Vec3D[int]],
+    gap: Vec3D[int],  # pylint: disable=unused-argument
```
dodamih, line 403: Remove pylint directive

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -473,8 +484,10 @@ def _expand_bbox_processing(  # pylint: disable=line-too-long
             f"{dst_resolution.pformat()}. You may set `expand_bbox_resolution = True` to "
             "automatically expand the bbox to the nearest integral pixel."
         )
-    translation_end = (processing_chunk_sizes[0] - bbox_shape_in_res) % processing_chunk_sizes[0]
     bbox_old = bbox
+    chunk_size = processing_chunk_sizes[0]
+    # Round up the bbox so that (bbox - chunk size) is an even multiple of (chunk size + gap).
+    translation_end = (chunk_size - bbox_shape_in_res) % (chunk_size + gap)
```
dodamih, line 490: Minor, but `chunk_size` is somewhat ambiguous - better to call this `chunk_size_top`.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -473,8 +484,10 @@ def _expand_bbox_processing(  # pylint: disable=line-too-long
...
     bbox = bbox.translated_end(translation_end, dst_resolution)
     if translation_end != Vec3D[int](0, 0, 0):
         logger.info(
```
dodamih, line 493: Commenting here since I can't select the correct line: the message on line 494 also needs to be updated to read:

f"`expand_bbox_processing` was set and the `bbox` was not aligned to the top level `processing_chunk_size` (with `gap`s if applicable) in at least one dimension, "

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -415,7 +424,8 @@ def make_tasks_with_intermediaries(  # pylint: disable=too-many-locals
         op_kwargs: P.kwargs,
     ) -> Tuple[List[mazepa.tasks.Task[R_co]], VolumetricBasedLayerProtocol | None]:
         dst_temp = self._get_temp_dst(dst, idx, self.flow_id)
-        idx_chunks = self.processing_chunker(idx, mode="exact")
+        have_gap = self.gap is not None and self.gap != Vec3D[int](0, 0, 0)
+        idx_chunks = self.processing_chunker(idx, mode="expand" if have_gap else "exact")
```
dodamih, line 428: What happens if it's `exact`? If the arguments are correct, then this should never need to be `expand`. Can you check the `idx_chunks` when it's `exact` and compare them to `expand`?

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -342,9 +344,16 @@ def __attrs_post_init__(self):
...
+        if self.gap is not None and self.gap != Vec3D[int](0, 0, 0):
+            self.processing_chunker = VolumetricIndexChunker(
+                chunk_size=self.processing_chunk_size,
+                resolution=self.dst_resolution,
+                stride=self.processing_chunk_size + self.gap,
+            )
+        else:
+            self.processing_chunker = VolumetricIndexChunker(
+                chunk_size=self.processing_chunk_size, resolution=self.dst_resolution
+            )
```
dodamih, line 356: We don't want to have `is not None` checks everywhere - see what's done for `self.roi_crop_pad`, `self.processing_blend_pad`, etc above: `self.gap` can just be set to `Vec3D[int](0, 0, 0)`, and then this if-else block can be removed.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -577,11 +587,22 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
...
+                )
+            else:
+                stride_and_chunk_size = (
+                    dst.backend.get_chunk_size(self.dst_resolution) - self.gap // 2
+                )
+                reduction_chunker = VolumetricIndexChunker(
+                    chunk_size=stride_and_chunk_size,
+                    resolution=self.dst_resolution,
+                    max_superchunk_size=stride_and_chunk_size,
+                )
```
dodamih, line 605: Same as above - I think we can get rid of this if-else block. Also: 
- `max_superchunk_size` should always be `copy_chunk_size`.
- `offset` should be set to `- self.gap // 2`.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
JoeStrout: Had to remove the whole line because, in fact, the argument was unused!

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
JoeStrout: If it's `exact` then code further on (in bbox_strider.py) throws this exception:

```
NotImplementedError: `exact` mode is only supported when the `chunk_size` and `stride` are equal
```

And of course they're _not_ equal here; the whole point of this is to specify a stride bigger than the chunk, so it leaves gaps.

Are you suggesting that we should update bbox_strider so it can deal with this case?

dodamih: I'll make an issue.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -877,6 +908,7 @@ def _build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg
         roi_crop_pad=roi_crop_pads[-1],
         processing_blend_pad=processing_blend_pads[-1],
         processing_blend_mode=processing_blend_modes[-1],
+        gap=None,  # no gap at the lowest level; only at the top level
```
dodamih, line 911: Here as well

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -473,12 +489,15 @@ def _expand_bbox_processing(  # pylint: disable=line-too-long
             f"{dst_resolution.pformat()}. You may set `expand_bbox_resolution = True` to "
             "automatically expand the bbox to the nearest integral pixel."
         )
-    translation_end = (processing_chunk_sizes[0] - bbox_shape_in_res) % processing_chunk_sizes[0]
     bbox_old = bbox
+    chunk_size_top = processing_chunk_sizes[0]
+    # Round up the bbox so that (bbox - chunk size) is an even multiple of (chunk size + gap).
```
dodamih, line 494: Delete comment please

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -415,7 +422,8 @@ def make_tasks_with_intermediaries(  # pylint: disable=too-many-locals
         op_kwargs: P.kwargs,
     ) -> Tuple[List[mazepa.tasks.Task[R_co]], VolumetricBasedLayerProtocol | None]:
         dst_temp = self._get_temp_dst(dst, idx, self.flow_id)
-        idx_chunks = self.processing_chunker(idx, mode="exact")
+        have_gap = self.gap is not None and self.gap != Vec3D[int](0, 0, 0)
+        idx_chunks = self.processing_chunker(idx, mode="expand" if have_gap else "exact")
```
dodamih, line 426: Could you add the comment above this line please:
`# TODO: remove "expand"; see https://github.com/ZettaAI/zetta_utils/issues/648`

dodamih (approved):

PR #643, opened 2024-02-20
feat: add some doc strings; correct some minor errors in documentation

JoeStrout:


supersergiy (changes requested):
Thanks for doing this!

Could you change the comment style to the native reST consistent the rest of the codebase [[example](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/layer/volumetric/cloudvol/backend.py#L74-L90)]? This might be a good application for ChatGPT

supersergiy (approved):

PR #647, opened 2024-02-22
Add sharded precomputed support

trivoldus28:


trivoldus28:
Branch is getting old so I'll make it draft for now. Will update and write tests when I have some time to work on it.

PR #649, opened 2024-02-22
feat: chunk alignment for TSBackend

dodamih:
Added chunk alignment for TSBackend, and cleaned up tests for VolumetricBackend / Precomputed.

Also added an overload for `round(Vec3D)` to type / make it behave consistently with Python's built-in round, alongside coverage for Vec3D.__init__ from numpy.


trivoldus28 (approved):
lgtm, thanks.

zetta_utils/layer/volumetric/tensorstore/backend.py
```diff
@@ -51,7 +51,11 @@ def _get_ts_at_resolution(
     return result
 
 
-def _clear_ts_cache(path: str) -> None:
+def _clear_ts_cache(path: str | None = None) -> None:  # pragma: no cover
```
supersergiy, line 54: Hella logic here, but yeah testing cache is a pain

supersergiy (approved):

PR #650, opened 2024-02-22
Add all-in-one pairwise alignment wrapper

trivoldus28:
Example CUE that I actually used provided. Generally ready for review, except mypy and I'm not sure if it's worth it? Also looking for feedback on how things can be different for usability.


trivoldus28:
Since it's likely to need significant work to get it pass tests, I'll make it draft for now and revisit it later next time I use it.

PR #654, opened 2024-02-26
feat: db support for collaborative annotations management

akhileshh:


zetta_utils/db_annotations/annotation.py
```diff
@@ -0,0 +1,157 @@
...
+def read_annotation(annotation_id: str) -> DBRowDataT:
+    idx = (annotation_id, INDEXED_COLS + NON_INDEXED_COLS)
+    return ANNOTATIONS_DB[idx]
+
+
+def read_annotations(
+    annotation_ids: list[str] | None = None,
+    collection_ids: list[str] | None = None,
+    layer_group_ids: list[str] | None = None,
+    tags: list[str] | None = None,
```
supersergiy, line 51: Seems like not all of these can have non-default values at the same time, `@overload` would be good here.

zetta_utils/db_annotations/annotation.py
```diff
@@ -0,0 +1,157 @@
...
+    AnnotationKeys.RADII.value,
+)
+
+DB_NAME = "annotations"
+DB_BACKEND = DatastoreBackend(DB_NAME, project=constants.PROJECT, database=constants.DATABASE)
+DB_BACKEND.exclude_from_indexes = NON_INDEXED_COLS
+ANNOTATIONS_DB = build_db_layer(DB_BACKEND)
+
+
+def read_annotation(annotation_id: str) -> DBRowDataT:
```
supersergiy, line 42: Do we want to give all these functions an optional `annotation_db` argument that defaults to `ANNOTATIONS_DB` to make it testable?

zetta_utils/db_annotations/collection.py
```diff
@@ -0,0 +1,55 @@
...
+DB_NAME = "collections"
+INDEXED_COLS = ("name", "created_by", "created_at", "modified_by", "modified_at")
+NON_INDEXED_COLS = ("comment",)
+
+DB_BACKEND = DatastoreBackend(DB_NAME, project=constants.PROJECT, database=constants.DATABASE)
+DB_BACKEND.exclude_from_indexes = NON_INDEXED_COLS
+COLLECTIONS_DB = build_db_layer(DB_BACKEND)
+
+
+def read_collection(collection_id: str) -> DBRowDataT:
```
supersergiy, line 22: Might be possible to have a more narrow type here, but for mypy to not complain we'll have to assert that the indexed item is indeed conforming to the more narrow type, or just use `cast`

zetta_utils/db_annotations/layer_group.py
```diff
@@ -0,0 +1,64 @@
...
+
+import uuid
+
+from zetta_utils.layer.db_layer import DBRowDataT, build_db_layer
+from zetta_utils.layer.db_layer.datastore import DatastoreBackend
+
+from . import constants
+
+DB_NAME = "layer_groups"
+INDEXED_COLS = ("name", "layers", "created_by", "modified_by")
```
supersergiy, line 13: Would it make more sense to keep a layer group per collection? So that the name for the same layers can be different depending on collection

supersergiy (approved):
Great work!

Issue #646, opened 2024-02-21
Subchunkable: blend mode "defer"


JoeStrout:
## Motivation

Sometimes we have work that can't be simply blended across chunk boundaries (for example, segmentation).  In this case, the user wants to specify blend_pads as usual, but defer the usual blending, and have as their output the intermediate checkerboard layers.  They can then run those 4 or 8 layers through whatever bespoke or more-complex code they need to produce one unified, coherent output.

## API Changes

In `build_subchunkable_apply_flow`, the `processing_blend_modes` parameter currently accepts "linear" or "quadratic".  We will add a third mode, "defer".  When this mode is selected:

1. Destination layer (`dst`) must _not_ be `None`
2. `skip_intermediaries` must be `False`
3. Files will be left in `level_intermediaries_dirs[-1]`
4. Those files will use a more human-readable naming scheme (skipping the usual hashes used to avoid conflicts when multiple jobs are sharing the same intermediary directory).  These names will favor underscores and avoid slashes, spaces, and parentheses, to make them easier to use on the command line.

As a related task that came up during implementation:

5. The three _procs properties (write_procs etc.), currently defined in `VolumetricLayer`, should be added to `VolumetricBasedLayerProtocol`.  This simplifies type-checking and mypy-wrestling, as it means we can count on anything implementing that protocol having these properties.

dodamih:
Oh, I also just noticed - `level_intermediaries_dirs` is largest first, so it'd actually be `level_intermediaries_dirs[0]`.

PR #655, opened 2024-02-26
feat: blend_mode = "defer"

JoeStrout:
https://github.com/ZettaAI/zetta_utils/issues/646

### DIFF
--- a/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
+++ b/zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
@@ -26,7 +26,11 @@ from typing_extensions import ParamSpec
 from zetta_utils import builder, log, mazepa
 from zetta_utils.common.pprint import lrpad, utcnow_ISO8601
 from zetta_utils.geometry import BBox3D, Vec3D
-from zetta_utils.layer.volumetric import VolumetricBasedLayerProtocol, VolumetricIndex
+from zetta_utils.layer.volumetric import (
+    VolumetricBasedLayerProtocol,
+    VolumetricIndex,
+    VolumetricLayer,
+)
 from zetta_utils.layer.volumetric.cloudvol.build import build_cv_layer
 from zetta_utils.layer.volumetric.tensorstore.build import build_ts_layer
 from zetta_utils.mazepa import SemaphoreType, id_generation
@@ -85,7 +89,7 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
     processing_crop_pads: Sequence[int] | Sequence[Sequence[int]] = (0, 0, 0),
     processing_blend_pads: Sequence[int] | Sequence[Sequence[int]] = (0, 0, 0),
     processing_blend_modes: Union[
-        Literal["linear", "quadratic"], Sequence[Literal["linear", "quadratic"]]
+        Literal["linear", "quadratic", "defer"], Sequence[Literal["linear", "quadratic", "defer"]]
     ] = "quadratic",
     level_intermediaries_dirs: Sequence[str | None] | None = None,
     skip_intermediaries: bool = False,
@@ -136,7 +140,8 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
         than or equal to half of the ``processing_chunk_size`` in each dimension.
     :param processing_blend_modes: Which blend mode to use at each subchunking level. ``linear``
         sums the blended areas weighted linearly by the position. ``quadaratic`` sums the
-        blended areas weighted quadratically by the position.
+        blended areas weighted quadratically by the position.  ``defer`` skips the final
+        reduction stage, leaving the final intermediate files for the user to handle.
     :param max_reduction_chunk_sizes: The upper bounds of the sizes chunks to be used for the
         reduction step. During the reduction step, backend chunks in the area to be reduced will
         be reduced in larger chunks that have been combined up to this limit. Reduction chunks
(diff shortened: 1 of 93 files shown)


zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -214,12 +219,33 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
             )
         bbox_ = bbox
 
+    defer_blend = processing_blend_modes == "defer" or processing_blend_modes[-1] == "defer"
+
     if dst is None:
+        if defer_blend:
+            raise ValueError('`dst` must be None when `processing_blend_modes` is "defer".')
```
dodamih, line 226: "cannot be None"

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -214,12 +219,33 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
...
 
+    if defer_blend:
+        logger.warning(
+            'Because `processing_blend_modes` is "defer", `dst` will NOT '
+            "contain final output of subchunkable."
+        )
+        if isinstance(dst, VolumetricLayer) and dst.write_procs:
+            logger.warning(
+                "Unblended intermediaries will already have the write_procs applied "
+                "to them; another task writing to `dst` may apply the write_procs a "
```
dodamih, line 245: "another Flow or task that writes to `dst`"

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -367,7 +367,7 @@ def _get_temp_dst(
         assert self.intermediaries_dir is not None
         temp_name = f"{prefix}_{self.op.__class__.__name__}_temp_{idx.pformat()}_{suffix}"
         allow_cache = self.allow_cache and not self._intermediaries_are_local
-        if self.use_checkerboarding:
+        if self.use_checkerboarding and dst is not None:
```
dodamih, line 370: I'm guessing this is a holdover from when we were okay with nonmandatory `dst`?

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -463,8 +463,9 @@ def make_tasks_with_checkerboarding(  # pylint: disable=too-many-locals, too-man
             backend chunk size to half of the processing chunk size. Furthermore, skip caching
             if the temporary destination happens to be local.
             """
-            assert dst is not None
-            dst_temp = self._get_temp_dst(dst, idx, self.flow_id, chunker_idx)
+            dst_temp = self._get_temp_dst(
+                dst, idx, self.flow_id, "_".join(str(i) for i in chunker_idx)
+            )
```
dodamih, line 468: Same as above

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -620,6 +621,43 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
...
+                dst_temps,
+            ) = self.make_tasks_with_checkerboarding(
+                idx.padded(self.roi_crop_pad), red_chunks, red_shape, dst, op_kwargs
+            )
+            logger.info(
+                "Writing to intermediate destinations:\n"
+                f" Submitting {len(tasks)} processing tasks from operation {self.op}."
+            )
+            yield tasks
+            yield mazepa.Dependency()
```
dodamih, line 660: The reduction chunks are completely useless in this case - the `make_tasks_with_checkerboarding` only needs them to keep track of which tasks are contained in which reduction chunks, and as we're not reducing we can avoid dealing with the reduction chunker and just pass in one "reduction chunk" that's the whole index alongside (1,1,1) the shape of the reduction chunks (which specifies how many reduction chunks are there in each dimension to do some indexing magic to figure indices out in constant time). So lines 626-658 can just be changed to:

```
            stride_start_offset = dst.backend.get_voxel_offset(self.dst_resolution)
            stride_start_offset_in_unit = stride_start_offset * self.dst_resolution
           (
                tasks,
                _,
                _,
                dst_temps,
            ) = self.make_tasks_with_checkerboarding(
                idx.padded(self.roi_crop_pad), [idx], Vec3D(1,1,1), dst, op_kwargs
            )
            logger.info(
                "Writing to intermediate destinations:\n"
                f" Submitting {len(tasks)} processing tasks from operation {self.op}.\n"
                f"Note that because blending is deferred, {dst.pformat()} will NOT "
                f"contain the final output."
            )
```
Note I'm using `dst.pformat()`, even though it's not part of the VolumetricLayerBasedProtocol at the moment (it is part of VolumetricLayer) - we should probably add it.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -620,6 +621,43 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
...
         # case with checkerboarding
```
dodamih, line 661: This should be moved to above the last case, and modified to `# cases with checkerboarding`

dodamih: (The comment on #572 should also be plural - my bad.)

dodamih (changes requested):

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
JoeStrout: Thanks, that's much neater.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
dodamih: Oh, I just meant the assert

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -214,12 +219,33 @@ def build_subchunkable_apply_flow(  # pylint: disable=keyword-arg-before-vararg,
...
     if dst is None:
+        if defer_blend:
+            raise ValueError('`dst` cannot be None when `processing_blend_modes` is "defer".')
         if not skip_intermediaries:
-            raise ValueError("`skip_intermediaries` must be True when `dst` is None.")
+            raise ValueError(
+                "`skip_intermediaries` must be True when `dst` is None and "
+                "`processing_blend_modes` is not "
+                "defer"
+                "."
```
dodamih, line 232: Combine lines? Linters do this sometimes....

JoeStrout: Yeah I definitely didn't do that!  Probably `black`'s fault.  Looks like it took out the intended quotation marks, too.  I'll see what I can do.

dodamih (approved):
Nice work!

PR #657, opened 2024-03-05
Slurm Integration

supersergiy:


nkemnitz (approved):
I only removed the output file, otherwise looks good. Thank you so much!

PR #664, opened 2024-04-15
feat: chunk_id in VolumetricIndex, with Subchunkable incrementing it

dodamih:


supersergiy (approved):
Preemptively approving for now because I'll be away until Wednesday and merging this in might be urgent for Tri's usecase

PR #665, opened 2024-04-16
fix(builder): lambda substitution preserves position

nkemnitz:
This fixes an issue encountered when using `read_procs` involving multiple steps and a `lambda` that is not the final proc. Previously, the `read_proc` would fail or produce incorrect results, because the `lambda` got substituted with an `invoke_lambda_str` placed at the end of the `read_procs`, regardless of its original position.

Solution feels a bit hacky, also because I don't fully understand why this positional switch happens and whether it applies to other bulding blocks, too. Open for suggestions.


supersergiy (approved):

supersergiy:
Found a better solution

PR #670, opened 2024-04-20
Slurm execution fixes

nkemnitz:
* use `srun` to actually run tasks in parallel
* improved error message when job creation fails
* allow disabling semaphore spec
* minor typing fix with gres


nkemnitz:
@supersergiy I removed the `execute_on_slurm_fq.py` - I think that one is now fully included in the regular `execute_on_slurm` version, correct?

supersergiy (approved):

PR #675, opened 2024-04-22
fix: builder list ordering

supersergiy:


zetta_utils/builder/building.py
```diff
@@ -164,11 +164,11 @@ def _process_result(obj: ObjectToBeBuilt, result: Any):
 
 
 def _build_list(**kwargs):
-    return list(kwargs.values())
+    return [value for key, value in sorted(kwargs.items())]
```
nkemnitz, line 167:
```suggestion
    return [value for _, value in sorted(kwargs.items(), key=lambda x: int(x[0]))]
```

zetta_utils/builder/building.py
```diff
@@ -164,11 +164,11 @@ def _process_result(obj: ObjectToBeBuilt, result: Any):
...
 
 
 def _build_tuple(**kwargs):  # pragma: no cover
-    return tuple(kwargs.values())
+    return tuple(value for key, value in sorted(kwargs.items()))
```
nkemnitz, line 171:
```suggestion
    return tuple(value for _, value in sorted(kwargs.items(), key=lambda x: int(x[0])))
```

nkemnitz (changes requested):
Oh, that's a better solution than mine, thanks!
Just one issue with the current version: The dict keys are strings, so read proc `'11'` would be handled before `'2'`. Need to convert the keys to int for comparison.

nkemnitz:
Added the required `int` conversion and corresponding test

PR #678, opened 2024-04-24
chore: deprecate p39 docker builds

### LINKED ISSUE #676: Deprecate docker 3.9 default build
Now that we've dropped support for Python 3.9, we should remove it as a build option (currently it's the default) and add 3.11 instead.

supersergiy:
Fixes #676


trivoldus28 (approved):
lgtm

PR #679, opened 2024-04-24
Add watershed functions

trivoldus28:
This PR adds the watershed function, with options to use either `abiss` or `lsd` (distance transform then `mahotas`'s watershed).

* TODO
  * Add an example spec
  * (later): Add an actual test with output comparison


supersergiy (approved):
The code looks great, good to merge once CI passes

PR #683, opened 2024-04-26
fix: float res `add_info_scale`

supersergiy:


trivoldus28 (approved):
lgtm, if there's really no compatibility issues with having float res in the pipeline

PR #693, opened 2024-05-07
Update internal submodule to latest commit on its main branch

JoeStrout:


trivoldus28 (approved):

supersergiy:
@trivoldus28 @JoeStrout 
The hiccup here is that we need to sync up the requirements between `pyproject.toml`'s when we update `internal` versions. It's not ideal, but I haven't found a way around it yet. We want both projects to be installable independently to run independent CI's, which drives the need for two pyprojects

trivoldus28:
I'm a bit confused, why does this repo need the extra dependencies if the tests and functionalities don't need them?

trivoldus28:
Maybe put a check if internal exists, install it. If not then don't?

supersergiy:
I'm guessing it's because neither the tests nor the rest of the CI actually import the file in question

PR #696, opened 2024-05-10
fix chunk_id in the case of blend padding

JoeStrout:


tests/integration/subchunkable/test_subchunkable.py
```diff
@@ -236,3 +237,48 @@ def test_subchunkable_chunk_ids(clear_temp_dir_and_info_cache):
...
+    "not config.getoption('--run-integration')",
+    reason="Only run when `--run-integration` is given",
+)
+def test_subchunkable_padded_chunk_ids(clear_temp_dir_and_info_cache):
+    global COLLECTED_CHUNK_IDS  # pylint: disable=global-statement
+    COLLECTED_CHUNK_IDS = []
+    dst_resolution = [16, 16, 42]
+    dst = build_cv_layer(
+        path="file://~/zetta_utils_temp/",
+        info_reference_path="gs://dkronauer-ant-001-synapse/test/inference20240329023545",
```
dodamih, line 252: This should point to something in assets, and edited to not need `info_fields_overrides` or `add_scales`.

dodamih (changes requested):
One comment about the info file (shouldn't point to a real dataset, but should be in assets).
Also please prefix commit messages with `fix:` and `chore:` respecstively.

tests/integration/subchunkable/test_subchunkable.py
dodamih: Also, the path should not be in home - you should make a scratch folder in assets.

dodamih (approved):

PR #697, opened 2024-05-13
feat: Added blend mode 'max'

JoeStrout:
This blends by taking the maximum value in the overlapping region.  Useful for things like segment IDs, where if a segment in the overlap has been labeled by both chunks, we just want all chunks to agree on which ID to use — and max is an efficient, consistent choice.

### DIFF
--- a/zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
+++ b/zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import itertools
 import multiprocessing
+from abc import ABC
 from copy import deepcopy
 from os import path
 from typing import Any, Generic, List, Literal, Optional, Tuple, TypeVar
@@ -50,7 +51,23 @@ class Copy:
 
 @mazepa.taskable_operation_cls
 @attrs.mutable
-class ReduceByWeightedSum:
+class ReduceOperation(ABC):
+    def __call__(
+        self,
+        src_idxs: List[VolumetricIndex],
+        src_layers: List[VolumetricBasedLayerProtocol],
+        red_idx: VolumetricIndex,
+        roi_idx: VolumetricIndex,
+        dst: VolumetricBasedLayerProtocol,
+        processing_blend_pad: Vec3D[int],
+        processing_blend_mode: Literal["linear", "quadratic"],
+    ) -> None:
+        pass
+
+
+@mazepa.taskable_operation_cls
+@attrs.mutable
+class ReduceByWeightedSum(ReduceOperation):
     def __call__(
         self,
         src_idxs: List[VolumetricIndex],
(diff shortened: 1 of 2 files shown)


zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -715,8 +766,13 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
             )
             yield tasks
             yield mazepa.Dependency()
+            reducer: ReduceOperation
+            if self.processing_blend_mode == "max":
+                reducer = ReduceByMaxValue()
+            else:
+                reducer = ReduceByWeightedSum()
```
dodamih, line 773: Change to one line:
`reducer = ReduceByMaxValue() if self.processing_blend_mode =="max" else ReduceByWeightedSum()`? I guess it might be a little long.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -107,6 +124,40 @@ def __call__(
...
+class ReduceByMaxValue(ReduceOperation):
+    def __call__(
+        self,
+        src_idxs: List[VolumetricIndex],
+        src_layers: List[VolumetricBasedLayerProtocol],
+        red_idx: VolumetricIndex,
+        roi_idx: VolumetricIndex,
+        dst: VolumetricBasedLayerProtocol,
+        processing_blend_pad: Vec3D[int],
+        processing_blend_mode: Literal["linear", "quadratic"],
```
dodamih, line 138: This should be `Literal["max"]`, and we might even want an assert right below.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -50,7 +51,23 @@ def __call__(
...
+class ReduceOperation(ABC):
+    def __call__(
+        self,
+        src_idxs: List[VolumetricIndex],
+        src_layers: List[VolumetricBasedLayerProtocol],
+        red_idx: VolumetricIndex,
+        roi_idx: VolumetricIndex,
+        dst: VolumetricBasedLayerProtocol,
+        processing_blend_pad: Vec3D[int],
+        processing_blend_mode: Literal["linear", "quadratic"],
```
dodamih, line 63: This should be `Union[Literal["linear", "quadratic"], Literal["max"]]`

dodamih (changes requested):
Mostly good, but needs some typing changes, and also needs tests.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -107,6 +124,40 @@ def __call__(
                 dst[red_idx] = res
 
 
+@mazepa.taskable_operation_cls
+@attrs.mutable
+class ReduceByMaxValue(ReduceOperation):
```
supersergiy, line 129: Let's name it `ReduceNaive` and have it only support `max` blend mode

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
supersergiy: It feels wrong to have the ABC have such concrete information about the classes that implement it

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
JoeStrout: Yeah I had it that way first, but one of our commit checks forced the line break, at which point I thought the version above was cleaner.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -50,7 +51,68 @@ def __call__(
...
+                    intscn, subidx = src_idx.get_intersection_and_subindex(red_idx)
+                    subidx_channels = [slice(0, res.shape[0])] + list(subidx)
+                    with semaphore("read"):
+                        res[subidx_channels] = layer[intscn]
+            with semaphore("write"):
+                dst[red_idx] = res
+
+
+@mazepa.taskable_operation_cls
+@attrs.mutable
```
dodamih, line 106: Should be `@attrs.frozen`.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -50,7 +51,68 @@ def __call__(
 
 @mazepa.taskable_operation_cls
 @attrs.mutable
```
dodamih, line 53: Should be deleted.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -50,7 +51,68 @@ def __call__(
...
+        red_idx: VolumetricIndex,
+        roi_idx: VolumetricIndex,
+        dst: VolumetricBasedLayerProtocol,
+        processing_blend_pad: Vec3D[int],
+    ) -> None:
+        pass
+
+
+@mazepa.taskable_operation_cls
+@attrs.mutable
```
dodamih, line 71: Should be `@attrs.frozen`.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
JoeStrout: Sorry, what should be deleted?  (These decorations are needed, or mypy freaks out.)

JoeStrout: Oh nvm — the first one is needed, but @attrs.mutable is not, that must be what you meant.

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
JoeStrout: OK.  But next week maybe somebody can explain to me why the (pre-existing) Copy operation on line 39 and ReduceByWeightedSum on line 107 need @attrs.mutable, but these new ones don't.

supersergiy (approved):

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
supersergiy: I think all of them should be frozen strictly speaking, but it doesn't matter too much. I'll just merge

PR #703, opened 2024-05-19
dblayer deletion, query, batching

akhileshh:


zetta_utils/layer/db_layer/datastore/backend.py
```diff
@@ -175,6 +157,67 @@ def write(self, idx: DBIndex, data: DBDataT):
...
+        wait=wait_exponential(multiplier=1, min=4, max=10),
+    )
+    def clear(self, idx: DBIndex | None = None):
+        if idx:
+            keys: list[Key] = []
+            for _key, col_keys in self._get_row_col_keys(idx.row_keys, str_keys=False).items():
+                keys.extend(col_keys)
+                keys.append(_key)
+            self.client.delete_multi(keys=keys)
+            return
```
supersergiy, line 184: Nag: I think `if/else` is slightly preferable to `if-return` in terms of readability, as it makes the exclusivity of the two options visually explicit

supersergiy (approved):
💯

PR #710, opened 2024-05-25
fix(gc): delete k8s job and configmap resources; look in default clus…

akhileshh:
…ter if no clusters associated with run; register clusterr for training


supersergiy (approved):
💯

PR #711, opened 2024-05-29
feat: add pad_center_to() wrapping torch.pad()

trivoldus28:
Pad a tensor to specified shape. This is a prereq for something else in `internal`


supersergiy (approved):
💯

PR #712, opened 2024-05-29
Joe/notebooks

JoeStrout:
Add some Jupyter notebooks I've been working on.


JoeStrout:
Dangit, looks like I screwed up somehow.

PR #717, opened 2024-06-09
feat(db_layer): adds firestore backend

akhileshh:


zetta_utils/constants.py
```diff
@@ -1,3 +1,4 @@
 """Common project-wide constants."""
 
 DEFAULT_PROJECT = "zetta-research"
+RUN_DATABASE: str | None = "run-db"
```
supersergiy, line 4: Why is it allowed to be `None`?

supersergiy (approved):

zetta_utils/constants.py
akhileshh: The emulator image used for firestore doesn't support custom database name, so we have to set it to `None` to indicate default database in tests.

PR #722, opened 2024-06-13
chore: add DBLayer to import for inference, update internal

dodamih:


dodamih:
@supersergiy I think this is a mypy bug - the operation is definitely a `VolumetricOpProtocol`, but it might not be recognised as such because of the `@mazepa.taskable_operation_cls` adding `make_task` rather than it being defined explicitly.

supersergiy:
Yep, feel free to `type: ignore` it, it's a known issue: https://github.com/ZettaAI/internal/blob/6c081ecc8a19ab39cb6531a89706300bab35a3a7/alignment/flows/compute_field_flow.py#L175

supersergiy (approved):

PR #733, opened 2024-06-30
fix: supports_dict typing

supersergiy:


zetta_utils/augmentations/imgaug.py
```diff
@@ -58,21 +58,14 @@ def _ensure_cxyn(x: Sequence[Tensor], ref: Sequence[TensorTypeVar]) -> list[Tens
     ...
 
 
-def _ensure_cxyn(
-    x: Tensor | Sequence[Tensor], ref: TensorTypeVar | Sequence[TensorTypeVar]
-) -> TensorTypeVar | list[TensorTypeVar]:
+def _ensure_cxyn(x, ref):
```
nkemnitz, line 61: Oh, I should only type the @overload versions?

nkemnitz (approved):

zetta_utils/augmentations/imgaug.py
supersergiy: yeah, that's the proper behavior as per official docs [https://docs.python.org/3/library/typing.html]
although sometimes VSCode encourages incorrect behavior by working better with explicit annotation

PR #734, opened 2024-06-30
feat: extend info mode

supersergiy:


JoeStrout (changes requested):
All looks good except for those two error messages.

PR #735, opened 2024-06-30
chore: reduce typechecked warnings

supersergiy:


zetta_utils/tensor_ops/convert.py
```diff
@@ -96,7 +96,7 @@ def astype(
 
     """
     if isinstance(reference, torch.Tensor):
-        data_tc = tensor_ops.convert.to_torch(data, device=reference.device)
+        data_tc = tensor_ops.convert.to_torch(data)
```
nkemnitz, line 99: I think I just added that in #721

nkemnitz (changes requested):
See comment regarding `convert.astype`

nkemnitz (approved):

PR #738, opened 2024-07-03
fix(dockerfile.all.p311): Make onnxruntime-gpu use CUDA

nkemnitz:


supersergiy (approved):

nkemnitz:
Actually not so minor size reduction. Looks like that turning off cache shaved off 2.5 GB of the compressed image! (5.5 GiB instead of 8 GiB)

PR #743, opened 2024-07-11
feat(cloudvol.backend): ScaleUnavailableError with path

nkemnitz:
Just adds the actual cloudpath that was causing the `ScaleUnavailableError`

### DIFF
--- a/zetta_utils/layer/volumetric/cloudvol/backend.py
+++ b/zetta_utils/layer/volumetric/cloudvol/backend.py
@@ -9,6 +9,7 @@ import cachetools
 import cloudvolume as cv
 import numpy as np
 from cloudvolume import CloudVolume
+from cloudvolume.exceptions import ScaleUnavailableError
 from numpy import typing as npt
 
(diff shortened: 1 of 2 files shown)


zetta_utils/layer/volumetric/cloudvol/backend.py
```diff
@@ -38,14 +39,17 @@ def _get_cv_cached(
...
+            result = CloudVolume(
+                path_,
+                info=get_info(path_),
+                provenance={},
+                mip=tuple(resolution),
+                lru_bytes=cache_bytes_limit,
+                **kwargs,
+            )
+        except ScaleUnavailableError as e:
+            raise ScaleUnavailableError(f"{path_} - {e}") from e
```
supersergiy, line 52: I think it needs either a test or `#pragma: no cover`

nkemnitz: Test added.

dodamih (approved):

PR #744, opened 2024-07-11
feat(generators): get_field_from_matrix

nkemnitz:
Similar to `torchfields.affine_field`, but supports generating fields represented by a perspective 3x3 2D transform


trivoldus28 (approved):
lgtm!

PR #746, opened 2024-07-12
Disable CDN cache for volatile JSON objects

nkemnitz:
I tend to modify info files more frequently during development and it always caused trouble. CloudVolume's `commit_info` hardcoded this as well to `no-cache`.

Also found a similar `put_json` call for the remote annotations, which I reckon act very much the same.

### DIFF
--- a/zetta_utils/parsing/ngl_state.py
+++ b/zetta_utils/parsing/ngl_state.py
@@ -115,4 +115,6 @@ def write_remote_annotations(
 
     cf = CloudFiles(remote_path)
     logger.info(f"Writing {len(bboxes_or_points)} bboxes/points to {remote_path}/{layer_name}.")
-    cf.put_json(layer_name, make_layer(layer_d).to_json())
+    cf.put_json(
+        layer_name, make_layer(layer_d).to_json(), compress="gzip", cache_control="no-cache"
+    )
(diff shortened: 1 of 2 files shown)


zetta_utils/parsing/ngl_state.py
```diff
@@ -115,4 +115,6 @@ def write_remote_annotations(
 
     cf = CloudFiles(remote_path)
     logger.info(f"Writing {len(bboxes_or_points)} bboxes/points to {remote_path}/{layer_name}.")
-    cf.put_json(layer_name, make_layer(layer_d).to_json())
+    cf.put_json(
+        layer_name, make_layer(layer_d).to_json(), compress="gzip", cache_control="no-cache"
```
akhileshh, line 119: Does this affect `get_json` in `read_remote_annotations` ? No need to pass `compress` there?
Looks good otherwise.

akhileshh (approved):

zetta_utils/parsing/ngl_state.py
nkemnitz: Hm... should not cause issues if the remote path is object storage: the encoding will be stored as `Content-Encoding` metadata and by default [CloudFiles `get` will return the decoded result](https://github.com/seung-lab/cloud-files/blob/master/cloudfiles/cloudfiles.py#L430)

There _could_ be a problem if you are writing to the file system, maybe? There is no metadata, so CloudFiles will append a .gz suffix to the filename

PR #751, opened 2024-07-24
Change specs submodule from http:// to git://

trivoldus28:
To be consistent with `internal` and to make my life easier when cloning (http:// doesn't use key and github disabled password based authentication a long time ago).


supersergiy (approved):
Thnx

PR #755, opened 2024-07-26
feat(training/LayerDataset): add length override

trivoldus28:


supersergiy:
Would [[LoopSampler](https://github.com/ZettaAI/zetta_utils/blob/e62d5502e92f4b123611c9c40cff056f5f47acfa/zetta_utils/training/datasets/sample_indexers/loop_indexer.py#L14)] work for this?

PR #756, opened 2024-07-26
feat(training/ROIMaskProcessor): add upstream_pad

trivoldus28:
Also change output mask type to uint8 by default

### DIFF
--- a/tests/unit/layer/volumetric/test_tools.py
+++ b/tests/unit/layer/volumetric/test_tools.py
@@ -151,7 +151,7 @@ def test_volumetric_index_scaler_error():
 
 
 @pytest.mark.parametrize(
-    "start_coord, end_coord, resolution, targets, data_shape, expected_mask_region, existing_masks",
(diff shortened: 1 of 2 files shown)


supersergiy (approved):
Looks good! Could you rebase instead of making merge commits next time? otherwise we get a lot of them and they can clutter history in some situations.

trivoldus28:
But I think maybe there should be a more long term solution. In the current setup I don't think we can have more than 2 augments that dynamically modify the index. Maybe save the original index in the current index that the processors can compare?

trivoldus28:
Some errors about docs this time.. I'll wait for that to be resolved first?

supersergiy:
Yup I'm on it

supersergiy:
> But I think maybe there should be a more long term solution. In the current setup I don't think we can have more than 2 augments that dynamically modify the index. Maybe save the original index in the current index that the processors can compare?

Yeah we should think about what to do...

PR #761, opened 2024-08-02
fix/typing(layer_set): Allow layer set read_proc to return single layer

nkemnitz:
Allows combining two layers with a read_proc, e.g. https://github.com/ZettaAI/specs/blob/main/nico/inference/chen_s1/11_fine_align_pairwise.cue#L196-L223

Already worked before, just the static type checker would complain.

### DIFF
--- a/zetta_utils/layer/volumetric/layer_set/layer.py
+++ b/zetta_utils/layer/volumetric/layer_set/layer.py
@@ -5,7 +5,6 @@ from typing import Mapping, Union
 import attrs
 import torch
 from numpy import typing as npt
-from typeguard import typechecked
 
 from ... import DataProcessor, IndexProcessor, JointIndexDataProcessor, Layer
 from .. import UserVolumetricIndex, VolumetricFrontend, VolumetricIndex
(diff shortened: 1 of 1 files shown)


nkemnitz:
Tests fail to unrelated `zetta_utils/internal/segmentation/embedding/utils.py:117: error: Argument 1 to "_compute_affinity" has incompatible type "ndarray[Any, dtype[Any]]"; expected "Tensor"  [arg-type]`

supersergiy:
I think Ideally we would want to handle it in a way that would still keep the return type of layer set's `__getitem__` predictable. With this change, every time a user reads from a layer set in python they would have to assert the expected return type. Let me think about this a bit...

nkemnitz:
Any update here?

supersergiy:
The issue with the layer set return type being an unpredictable union is still not addressed. If a read processor is changing the type, it is in fact a type bug. A read processor can change the type to literally anything, which means that a cast or type ignore is waranted

nkemnitz:
New attempt and some sort of middle ground until we maybe find a better solution or collect more use cases to help decide:

* `cast` and `#type: ignore` I cannot use, because the layer set is created from within the cue spec.
* building an entire new `ReducedVolumetricLayerSet` class only for this special use case that one of my read_procs is reducing the dictionary feels too bloated.
* removing the dynamic type checker keeps mypy functionality and related errors/warnings, but still allows more powerful read procs, if necessary.

supersergiy (approved):

PR #775, opened 2024-08-14
Minor BBox3D/VolumetricIndex improvements

JoeStrout:
1. Added `from_points`, `union`, and `contains` methods to BBox3D.
2. Clarified (and corrected minor typos in) doc strings in BBox3D.
3. In VolumetricIndex, modified `from_coords` so it take its resolution as a CUE-friendly sequence as well as a Vec3D; also registered it with the builder, so you can actually create one in a CUE file now.
4. Added `union` and `contains` methods to VolumetricIndex as well.
5. Added doc strings to VolumetricIndex, which previously had none.

### DIFF
--- a/zetta_utils/geometry/bbox.py
+++ b/zetta_utils/geometry/bbox.py
@@ -24,9 +24,20 @@ Tuple2D = tuple[float, float]
 @typechecked
 class BBox3D:  # pylint: disable=too-many-public-methods # fundamental class
     """
-    3-Dimentional cuboid in space.
-    :param bounds: Bounds along X, Y, Z dimensions.
-    :param unit: Unit name (for decorative purposes only).
+    3-Dimensional axis-aligned cuboid in space.
+
+    By convention, values in a BBox3D always represent nm (the "unit" property
+    used only as a sanity check in methods like `intersects`).  In particular,
+    do not use a BBox3D to represent a bounding box in voxel space; see
+    VolumetricIndex for that.  When a BBox3D is created with a 'resolution'
+    argument, the values of the other arguments are immediately multiplied by
+    resolution, resulting in values in nm.  To convert from a BBox3D (in nm)
+    back to voxels at some resolution, you can use get_slice or from_slices.
+
+    :param bounds: (Min, max) values along X, Y, Z dimensions.
+    :param unit: Unit name (for decorative or validation purposes only).
+    :param pprint_px_resolution: Resolution used ONLY by the pformat method
+    (for pretty-printed coordinates suitable for use with Neuroglancer).
     """
 
     bounds: tuple[Tuple2D, Tuple2D, Tuple2D]
@@ -59,7 +70,7 @@ class BBox3D:  # pylint: disable=too-many-public-methods # fundamental class
         resolution: Sequence[float] = (1, 1, 1),
         unit: str = DEFAULT_UNIT,
     ) -> BBox3D:
-        """Create a `BBoxND` from slices at the given resolution.
+        """Create a `BBox3D` from slices at the given resolution.
 
         :param slices: Tuple of slices representing a bounding box.
         :param resolution: Resolution at which the slices are given.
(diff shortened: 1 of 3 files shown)


zetta_utils/geometry/bbox.py
```diff
@@ -518,37 +567,67 @@ def get_size(self) -> Union[int, float]:  # pragma: no cover
...
             tuple[Tuple2D, Tuple2D, Tuple2D],
             tuple((max(s[0], o[0]), min(s[1], o[1])) for s, o in zip(self.bounds, other.bounds)),
         )
         return BBox3D(bounds=bounds, unit=self.unit)
 
+    def union(self: BBox3D, other: BBox3D) -> BBox3D:
+        """Returns the union of another BBox3D with this one.
+        The `unit` property must match, but pprint_px_resolution is ignored.
+        """
+        assert self.unit == other.unit
```
dodamih, line 609: If it really is the union (rather than the supremum), then there should be `assert self.is_aligned(other)`.

dodamih (changes requested):

zetta_utils/geometry/bbox.py
JoeStrout: This comment really made me think!  I've always seen this operation (finding the smallest bounding box that contains both of the operand bounding boxes) called "union," but you're right, technically it's not a union in the sense of a non-box-shaped union of two arbitrary regions.  The area of the result here could be dramatically more than the sum of the areas of the inputs.

So I did some searching, but I can't find any references that don't call this operation on axis-aligned bounding boxes either "union" or (more rarely) "merge."  And I can't find any that apply the term "supremum" to bounding boxes at all — that generally means just the max of a set, which in this case I guess would be the point that's an upper bound for both bounding boxes.  I guess that's one corner of the result, with the other set to the infimum, but it still doesn't seem like a good term for what we're doing here.

I even went so far as to [ask ChatGPT](https://chatgpt.com/share/62d6e832-f218-40a3-9228-05d17c37b1a3) about it, figuring it has a pretty good grip on typical use of language.  It agrees this is a union operation (though also suggested merge).

So... how about we either leave this alone (as I think "union" is more standard), or call it `merge`, and in either case we make clear in the docstring that this is finding the minimum BBox3D that encloses both boxes?

dodamih: I still think it should be changed to `supremum` - it might also be good to make it a function that takes an iterable rather than a method.

So "supremum" generally refers to the least upper bound of a partially ordered (binary relation that's transitive and reflexive, not necessarily total) set, rather than the maximum. If you consider the sets in a topological space partially ordered under inclusion, then the supremum of two sets would be the union; for a set of numbers partially (and totally) ordered under the usual order, the supremum is the max of a set; for rectilinear bounding boxes partially ordered under inclusion, the supremum is the result of this operation. (Category theoretically, this is also called the "pushout" of two objects, which is dual to the "pullback" that one might encounter in other contexts.)

ChatGPT assumes that "supremum" given without context is in the category of sets of real numbers, since it probably has a lot more material on introductory real analysis than set theory, but you can get it to print the correct answer: 
Screenshot from 2024-08-22 15-15-00

JoeStrout: Okey dokey; renamed to supremum.  I left it as a method to match the interface of `intersection`, and mentioned "union" in the docstring in case somebody goes looking for it with that term.

dodamih: I'll approve - though mypy is failing at the moment with the name change.

dodamih (approved):

zetta_utils/layer/volumetric/index.py
```diff
@@ -162,3 +283,28 @@ def snapped(
             chunk_id=self.chunk_id,
             allow_slice_rounding=self.allow_slice_rounding,
         )
+
+    def union(self: VolumetricIndex, other: VolumetricIndex) -> VolumetricIndex:
```
dodamih, line 287: Also needs renaming and docstring update.

dodamih (changes requested):
Couple fixes needed

zetta_utils/geometry/bbox.py
JoeStrout: I can tell you that in the game industry, at least, this operation is very common and is universally called union (whether it should be or not).

But I don't want to argue about it; as long as the word is in there, it'll be found when somebody searches for it, and if they're confused by this comment, they can talk to you about it. :)  Wording updated.

PR #777, opened 2024-08-22
fix: use names as ids for convenience

akhileshh:


supersergiy (approved):
Thanks for the quick fix!

PR #784, opened 2024-08-29
feat: mazepa hooks for AnnotationLayer setup and post-process; build_annotation_layer builder

JoeStrout:
These changes enable the synapse detection and assignment pipeline to be run entirely from a CUE file, in principle making it usable from Zetta Portal.

(This is a high priority for @macrintr.)


JoeStrout:
Note: I've also got changes to internal and specs.  Should we merge the [internal PR](https://github.com/ZettaAI/internal/pull/48) first, and then include updated pointers to internal and specs in this PR?  Or update the pointers later in their own mini-PR?

supersergiy (approved):

PR #786, opened 2024-08-30
HTTP/2 + Streaming Response

nkemnitz:
Based on the comments in https://cloud.google.com/run/quotas#request_limits
* Remove 32 MiB request size limit by switching from uvicorn to hypercorn with HTTP/2 support
* Remove 32 MIB response size limit by using FastAPI Streaming response (which sets `Transfer-Encoding: chunked`)


nkemnitz:
Oh, no - wait. Still need to do something to get a StreamingResponse....

akhileshh (approved):
Thanks!

PR #793, opened 2024-09-10
chore: tools for np-centric execution

supersergiy:


zetta_utils/builder/__init__.py
```diff
@@ -11,3 +13,8 @@
 from . import built_in_registrations
 
 PARALLEL_BUILD_ALLOWED: bool = False
+
+NP_MANUAL: List[str] = []
+for k in dir(np):
```
nkemnitz, line 18: Numpy methods are a bit strange... it's a wild mix of `np._ArrayFunctionDispatcher`, `np.ufunc` and regular methods, so we can't use the same trick that worked for [ torch ](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/convnet/architecture/primitives.py#L28)

Looking at https://github.com/numpy/numpy/issues/24019 the recommended way seems to be `inspect.isroutine`. I would use that as well as skip all `k.startswith('_')`.

You might also want to register methods from submodules `np.linalg` and `np.fft`, similar to how we did it for torch

zetta_utils/convnet/simple_inference_runner.py
```diff
@@ -15,15 +17,16 @@ class SimpleInferenceRunner:  # pragma: no cover
     model_path: str
     unsqueeze_to: int | None = None
 
-    def __call__(self, src: torch.Tensor) -> torch.Tensor:
+    def __call__(self, src: npt.NDArray) -> npt.NDArray:
```
nkemnitz, line 20: At least the input should be our generic `Tensor`, in case the input already is a `torch.Tensor`?

zetta_utils/convnet/simple_inference_runner.py
```diff
@@ -15,15 +17,16 @@ class SimpleInferenceRunner:  # pragma: no cover
...
             device = "cpu"
 
         # load model during the call _with caching_
         model = convnet.utils.load_model(self.model_path, device=device, use_cache=True)
+
         if self.unsqueeze_to is not None:
             src = tensor_ops.unsqueeze_to(src, self.unsqueeze_to)
-        result = model(src.to(device))
+        result = model(to_torch(src).to(device))
         return result
```
nkemnitz, line 32: Return type (`torch.Tensor`) conflicts with the new return type definition (`npt.NDArray`).  We are a bit inconsistent in our code base. E.g. the `ModelInferencer` converts the result to the input data type, but in other locations we just always return `torch.Tensor`...

nkemnitz (changes requested):

nkemnitz (approved):

Issue #797, opened 2024-09-18
add /precomputed_annotations endpoint to web API


JoeStrout:
Portal needs a way to read and write a set of line annotations in precomputed format.  So:

Add a `/precomputed_annotations` endpoint to the web API, with two verbs:

GET: retrieves a set of line annotations, given these query parameters:

- `path`: GS path to the precomputed annotation file
- `bbox_start`: bounding box start (same format as `/painting/cutout`)
- `bbox_end`: bounding box start (same format as `/painting/cutout`)
- `resolution`: voxel size at which `bbox_start` and `bbox_end` are defined (defaults to 1,1,1)

Annotations are returned in the same format as `/annotations/multiple`.

PUT: replaces all line annotations within the given bounds with the new set of lines:

- `annotations`: new set of lines, in the same format as used by `/annotations/multiple`
- `path`: GS path to the precomputed annotation file
- `bbox_start`: bounding box start (same format as `/painting/cutout`)
- `bbox_end`: bounding box start (same format as `/painting/cutout`)
- `resolution`: voxel size at which `bbox_start` and `bbox_end` are defined (defaults to 1,1,1)

Note that with PUT, if the file does not exist, it will be created with the given bounds.  If it does exist, it will be updated, replacing only the lines within the given bounding box, but updating the overall bounds of the file to be the union of the old bounds and the given bounds.  _(Or, should we keep it simple and replace the entire file with the new bounds and lines?)_

supersergiy:
@JoeStrout should we close this?

PR #798, opened 2024-09-23
Precomputed_annotations web API (#797).

JoeStrout:
Implements #797.

For PUT and DELETE verbs, it turned out to be simplest to treat the file as a unit, i.e., PUT replaces the entire file and DELETE deletes the file.  (This also seems like a good fit for the resource model of REST APIs.)

Some changes also affect the `/annotations` endpoint, which @akhileshh found was broken in the current version.  It should be fixed with this PR.


akhileshh (approved):
Thanks!

PR #807, opened 2024-10-11
Joe/hotfix uint32 writes

JoeStrout:
Yesterday I found a bug where it was impossible to write to a CloudVolume of type `uint32` (at least when using interpolation type `segmentation`, because under the hood, even if you gave it data of that type, it would be converted to `int32` before writing.

This fixes that by casting back to the original data type after interpolation.


JoeStrout:
Slack discussion is [here](https://zettaai.slack.com/archives/C03TJ1FPJTD/p1728579755255949).

PR #811, opened 2024-10-17
Intersection support for queries; add created and modified timestamps to annotations.

akhileshh:


nkemnitz (approved):
Looks good, thanks for the update!

PR #818, opened 2024-10-28
chore: tools for np-centric execution

supersergiy:


nkemnitz:
Conflict needs to be fixed, but good to merge otherwise

nkemnitz (changes requested):
Conflict needs to be fixed, but good to merge otherwise

supersergiy:
Yup, looks like it! Thanks for pointing it out

PR #819, opened 2024-10-28
Performance optimization in precomp_annotations.subdivide

JoeStrout:
(substantially speeds up finalization of large annotation files)

### DIFF
--- a/zetta_utils/db_annotations/precomp_annotations.py
+++ b/zetta_utils/db_annotations/precomp_annotations.py
@@ -12,6 +12,7 @@ References:
 """
 
 import io
+import itertools
 import json
 import os
 import struct
@@ -145,7 +146,7 @@ def write_bytes(file_or_gs_path: str, data: bytes):
     :param file_or_gs_path: path to file to write (local or GCS path)
     :param data: bytes to write
     """
-    if not "//" in file_or_gs_path:
+    if "//" not in file_or_gs_path:
         file_or_gs_path = "file://" + file_or_gs_path
     cf = CloudFile(file_or_gs_path)
     cf.put(data)
@@ -184,6 +185,14 @@ def write_lines(file_or_gs_path: str, lines: Sequence[LineAnnotation], randomize
     write_bytes(file_or_gs_path, buffer.getvalue())
 
 
+def line_count_from_file_size(file_size: int) -> int:
+    """
+    Provide a count (or at least a very good estimate) of the number of lines in
+    a line chunk file of the given size in bytes.
+    """
+    return round((file_size - 8) / (LineAnnotation.BYTES_PER_ENTRY + 8))
+
+
 def count_lines_in_file(file_or_gs_path: str) -> int:
     """
     Provide a count (or at least a very good estimate) of the number of lines in
@@ -192,10 +201,7 @@ def count_lines_in_file(file_or_gs_path: str) -> int:
     # We could open the file and read the count in the first 8 bytes.
     # But even faster is to just calculate it from the file length.
     cf = CloudFile(file_or_gs_path)
-    fileLen = cf.size()
-    if fileLen is None:
-        return 0
-    return round((fileLen - 8) / (LineAnnotation.BYTES_PER_ENTRY + 8))
+    return line_count_from_file_size(cf.size() or 0)
 
 
 def read_bytes(file_or_gs_path: str):
@@ -205,7 +211,7 @@ def read_bytes(file_or_gs_path: str):
     :param file_or_gs_path: path to file to read (local or GCS path)
     :return: bytes read from the file
     """
-    if not "//" in file_or_gs_path:
+    if "//" not in file_or_gs_path:
         file_or_gs_path = "file://" + file_or_gs_path
     cf = CloudFile(file_or_gs_path)
     return cf.get()
@@ -470,7 +476,7 @@ class AnnotationLayer:
         """
         # Delete all files under our path
         path = self.path
-        if not "//" in path:
+        if "//" not in path:
             path = "file://" + path
         cf = CloudFiles(path)
         cf.delete(cf.list())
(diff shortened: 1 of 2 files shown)


zetta_utils/db_annotations/precomp_annotations.py
```diff
@@ -328,22 +329,45 @@ def subdivide(data, bounds: VolumetricIndex, chunk_sizes, write_to_dir=None, lev
...
+                for z in range(grid_shape[2]):
+                    qty_done += 1
+                    chunk_start = bounds.start + Vec3D(x, y, z) * chunk_size
+                    chunk_end = chunk_start + chunk_size
+                    chunk_bounds = VolumetricIndex.from_coords(
+                        chunk_start, chunk_end, bounds.resolution
+                    )
+                    # pylint: disable=cell-var-from-loop
+                    chunk_data: Sequence[LineAnnotation] = list(
+                        filter(lambda d: d.in_bounds(chunk_bounds), data_within_xy)
```
nkemnitz, line 361: I think passing chunk_bounds as default parameter might silence pylint. Not sure if that's cleaner, though
```suggestion
                        filter(lambda d, cb=chunk_bounds: d.in_bounds(cb), data_within_xy)
```

zetta_utils/db_annotations/precomp_annotations.py
```diff
@@ -593,16 +622,21 @@ def post_process(self):
           write_annotations(data, False), which writes to only the lowest-level chunks.
         """
         # read data (from lowest level chunks)
+        print("self.read_all")
```
nkemnitz, line 625: Replace print statements with logger calls (see other modules)

JoeStrout: Oops, the print statements were there for testing and I meant to remove them before committing.  Should be fixed now.

zetta_utils/db_annotations/precomp_annotations.py
JoeStrout: Oh actually I tried this, and mypy doesn't like it.  So I'm going to go back to the pylint disable approach.

nkemnitz:
Is this one ready for review?

JoeStrout:
Yep!

JoeStrout:
Hey @nkemnitz , I pushed some additional changes, but I now hope to quit mucking with this file for a while (and if I do, it'll be in a new branch).  Could you please review?

zetta_utils/db_annotations/precomp_annotations.py
```diff
@@ -592,14 +722,26 @@ def post_process(self):
...
+            # count in each one, and keep track of the max.
+            max_line_count = self.find_max_size(0)
+            print(f"Found max_line_count = {max_line_count}")
+            spatial_entries = self.get_spatial_entries(max_line_count)
+        else:
+            # Multiple chunk sizes means we have to start at the lowest
+            # level, and re-subdivide it at each higher level.
+
+            # read data (from lowest level chunks)
+            all_data = self.read_all()
```
nkemnitz, line 738: Same as for the size requests above - when loading annotations from object storage, you will get a considerable speedup by downloading all objects in one batch, rather than one by one as `read_all` / `read_lines` currently do.

zetta_utils/db_annotations/precomp_annotations.py
```diff
@@ -508,25 +550,91 @@ def write_annotations(self, annotations: Sequence[LineAnnotation], all_levels: b
...
+                    self.index.resolution,
                 )
                 # pylint: disable=cell-var-from-loop
-                chunk_data: list[LineAnnotation] = list(
-                    filter(lambda d: d.in_bounds(chunk_bounds), annotations)
+                split_data_by_x: list[LineAnnotation] = list(
+                    filter(lambda d: d.in_bounds(split_by_x), annotations)
                 )
-                if not chunk_data:
+                print(f":  {len(split_data_by_x)} lines")
```
nkemnitz, line 579: `logger.debug`, if worth keeping around

nkemnitz (commented):
I think you could simplify the code _and_ get a speed boost by using batch requests for fetching size and data using `CloudFiles`.

Other than that, the single chunk case doesn't seem to be covered by tests, yet.

zetta_utils/db_annotations/precomp_annotations.py
```diff
@@ -545,16 +653,38 @@ def read_all(self, spatial_level: int = -1, filter_duplicates: bool = True):
...
+        chunk_size = Vec3D(*self.chunk_sizes[level])
+        grid_shape = ceil(bounds_size / chunk_size)
+        level_key = f"spatial{level}"
+        level_dir = path_join(self.path, level_key)
+        result = 0
+        for x in range(0, grid_shape[0]):
+            for y in range(0, grid_shape[1]):
+                for z in range(0, grid_shape[2]):
+                    anno_file_path = path_join(level_dir, f"{x}_{y}_{z}")
+                    line_count = count_lines_in_file(anno_file_path)
```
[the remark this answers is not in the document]
JoeStrout: Great stuff, thank you!

zetta_utils/db_annotations/precomp_annotations.py
JoeStrout: That's a good point.  In this case, though, file I/O is not the bottleneck (even over the network), and refactoring it to download all at once would require a pretty major redesign.

zetta_utils/db_annotations/precomp_annotations.py
JoeStrout: It's definitely worth keeping — when this operation runs, it can take several hours, and it's very disconcerting (and inconvenient) to have no feedback during that time.  The current prints provide a nicer UX than logger, because I can control the carriage return.

BUT, I get it.  Logger is our standard way of corralling output, and works better when the code is called as part of a subchunkable flow.  So I'll change them all to logger.debug as suggested.

nkemnitz (approved):

PR #826, opened 2024-11-12
fix(cloudvol/write): allow sharded writes by removing forced autocrop

nkemnitz:
@supersergiy This autocrop was added two years ago. Can you imagine why? Right now it prevents me from uploading shards at the dataset boundary, when the shard as a whole is not aligned with the dataset bounds.


dodamih:
It's been a while since I've touched this part of the code, but I think the autocrop handles some weird cases when the user needs to process outside the edge of the dataset but only the internals get written. Does this branch pass integration tests?

nkemnitz:
Ah, probably not - I noticed CV complaining at me for different tasks with this change when my processing chunk size was larger than the dataset bounds.

nkemnitz:
Need to rethink how to handle sharded writes...

PR #832, opened 2024-12-05
fix: numerical stability on decimal resolutions

dodamih:
Fixes issues with floating point behaviour when using decimal resolutions.


nkemnitz (approved):
Thanks!

PR #833, opened 2024-12-13
Add test_custom_op.py to test RecursionError bug.

JoeStrout:
When this test was written (several months ago), it exposed a bug (threw a RecursionError exception, I think).

Now the test completes successfully, so yay!  Seems the bug was fixed.  But now we have a test for it, so that's good, right?


supersergiy (approved):
Looks good! Thanks for adding the test

PR #854, opened 2025-01-19
feat(docker): include pcg installation

akhileshh:


nkemnitz (changes requested):
In case you want to try this approach with the graph tool installation in the zutils image, I would request to cleanup the image from all temporary data, caches and unused apt packages.

@ranlu showed me how to do this for the Pytorch base image: (https://github.com/ZettaAI/zetta_utils/blob/main/docker/pytorch_base/Dockerfile#L122-L140)

akhileshh:
Thanks! But I think I like your pr better. I have no idea how long re-building the base image is going to take. Don't have the time to experiment with that now.

PR #856, opened 2025-01-21
feat(docker): Dockerfile with PCG support

nkemnitz:
@akhileshh Maybe like this?


akhileshh:
I like this better, and can confirm it works.

akhileshh (approved):

PR #858, opened 2025-01-22
Collection based dataset

supersergiy:
Build a dataset out of an annotation collection.

Annotation collection consists of annotations (shapes, boxes) and the associated layer groups. The collection based dataset will build a corresponding layer set and build a dataset.


nkemnitz (approved):

nkemnitz:
Very cool!

PR #863, opened 2025-01-23
Joe/feat 806

JoeStrout:
Adds `clearing_bbox` parameter to Annotation.write_in_bounds, and uses this in the PUT web API to replace only the annotations in the specified region.

Also shores up how resolution is handled — reading and writing now allow the caller to specify the resolution at which the line coordinates should be interpreted (converting to/from the underlying file resolution as needed).

### DIFF
--- a/web_api/app/precomputed_annotations.py
+++ b/web_api/app/precomputed_annotations.py
@@ -17,7 +17,7 @@ from zetta_utils.db_annotations.precomp_annotations import (
     AnnotationLayer,
     build_annotation_layer,
 )
-from zetta_utils.geometry import Vec3D
+from zetta_utils.geometry import BBox3D, Vec3D
 from zetta_utils.layer.volumetric import VolumetricIndex
 
 from .utils import generic_exception_handler
@@ -38,12 +38,14 @@ async def read_in_bounds(
     resolution: Annotated[tuple[float, float, float], Query()],
 ):
     """
-    This endpoint retrieves all lines within the given bounds.
+    This endpoint retrieves all lines entirely within the given bounds.
+    Coordinates are returned in units according to `resolution`.
     """
-    index = VolumetricIndex.from_coords(bbox_start, bbox_end, Vec3D(*resolution))
+    resolution_vec = Vec3D(*resolution)
+    bbox = BBox3D.from_coords(bbox_start, bbox_end, resolution_vec)
     layer = build_annotation_layer(path, mode="read")
     response = []
-    for line in layer.read_in_bounds(index):
+    for line in layer.read_in_bounds(bbox, strict=True, annotation_resolution=resolution_vec):
         annotation = AnnotationDBEntry(
             id=line.id,
             layer_group="",
@@ -65,8 +67,8 @@ async def add_multiple(
     resolution: Annotated[tuple[float, float, float], Query()],
 ):
     """
-    The PUT endpoint replaces all data in the given file (which may or
-    may not exist yet) with the given new set of lines.
+    The PUT endpoint replaces all data within the given bounds in the given file
+    (which may or may not exist yet) with the given new set of lines.
     """
     lines = []
     for entry in annotations:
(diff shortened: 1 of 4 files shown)


web_api/app/precomputed_annotations.py
```diff
@@ -38,12 +38,14 @@ async def read_in_bounds(
...
+    This endpoint retrieves all lines entirely within the given bounds.
+    Coordinates are returned in units according to `resolution`.
     """
-    index = VolumetricIndex.from_coords(bbox_start, bbox_end, Vec3D(*resolution))
+    resolution_vec = Vec3D(*resolution)
+    bbox = BBox3D.from_coords(bbox_start, bbox_end, resolution_vec)
     layer = build_annotation_layer(path, mode="read")
     response = []
-    for line in layer.read_in_bounds(index):
+    for line in layer.read_in_bounds(bbox, strict=True, annotation_resolution=resolution_vec):
```
supersergiy, line 48: Should we decouple the bbox specification resolution from the resolution we want the resulting annotations to be in?

web_api/app/precomputed_annotations.py
```diff
@@ -65,8 +67,8 @@ async def add_multiple(
     resolution: Annotated[tuple[float, float, float], Query()],
```
supersergiy, line 67: Same for the write -- should we allow the start and end coord to be given in resolutions that are not the same as the ones used for annotations?

zetta_utils/db_annotations/precomp_annotations.py
```diff
@@ -57,7 +58,9 @@ def __init__(self, line_id: int, start: Sequence[float], end: Sequence[float]):
         :param start: A tuple of three floats representing the start coordinate (x, y, z).
         :param end: A tuple of three floats representing the end coordinate (x, y, z).
 
-        Coordinates are in units defined by "dimensions" in the info file.
```
supersergiy, line 60: btw, is there a reason we're not using `attrs` here like we do in the rest of the codebase?

zetta_utils/internal
supersergiy, line 1: does the internal need to update here or is this an accident?

web_api/app/precomputed_annotations.py
JoeStrout: I'd say that's up to the users of the web API (your team in Ukraine, I guess).  Most of the time I would expect them to be the same; it's also easier to avoid mistakes if you don't have to keep two different resolutions in mind.  But if they have some reason to need them separate, we can easily do that.

web_api/app/precomputed_annotations.py
JoeStrout: Same answer.  I don't have strong feelings about it, but slightly prefer to have only one resolution.

zetta_utils/db_annotations/precomp_annotations.py
JoeStrout: Just that I'm still not super familiar with it, or the reasons one might want to use it.

zetta_utils/internal
JoeStrout: Dangit.  That's an accident again.  To check my code against the pre-commit hooks I have to commit them multiple times, and when there are multiple files, the easiest way to do that is `commit -a`.  And I forget that, even though I didn't touch internal, git sees it as changed (I guess because I was required, by another pre-commit hook, to switch internal to main).  So, that must be how this keeps happening.

I will try to undo that part.

web_api/app/precomputed_annotations.py
supersergiy: It's just a cleaner more modern way to use python that reduces boilerplate and comes with a lot of free gooddies

zetta_utils/db_annotations/precomp_annotations.py
supersergiy: It's just a cleaner more modern way to use python that reduces boilerplate and comes with a lot of free gooddies

supersergiy: We are trying to use it for all classes except for the ones that inherent from pytorch.nn.module (pytorch does some incompatible black magic)

zetta_utils/internal
JoeStrout: Oh!  Actually, no, in this case I do need to update internal as well (to track changes to the AnnotationLayer API).  I'll make a separate PR for that.

JoeStrout: https://github.com/ZettaAI/internal/pull/67
Not sure whether/how to remove the change to internal in this case, or make it point at the right thing.  But maybe it doesn't matter, since to get everything tidied up we will need to merge internal PR#67, and then pull that into main, and make another mini-PR on main to update internal, right?

zetta_utils/db_annotations/precomp_annotations.py
JoeStrout: OK.  I know you were eager to get this to your web team, so it's up to you whether you want to hold this until I can do that refactoring (probably tomorrow), or merge this as-is and I'll do another little PR to add `attrs`.

supersergiy (approved):

PR #865, opened 2025-01-24
build_image.py: `tag`->`tag_suffix`

supersergiy:
It's unintuitive that the actual docker tag is not identical to `--tag` argument given to the script. Too much know-how

### DIFF
--- a/build_image.py
+++ b/build_image.py
@@ -3,14 +3,20 @@
 import argparse
 import subprocess
 
-BUILD_COMMAND_TMPL = "docker build --network=host -t {REGION}-docker.pkg.dev/{PROJECT}/{REPO}/zetta_utils:py{PYTHON_VERSION}_{TAG} --build-arg PYTHON_VERSION={PYTHON_VERSION} -f docker/Dockerfile.{MODE} ."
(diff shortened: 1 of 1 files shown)


nkemnitz:
The python tag was supposed to make it easier catching issues from scheduler and image versions being out of sync. I was even thinking I should add _more_ tags for Pytorch and CUDA version, similar to how their official images are tagged.

Doesn't it show the full image path with name and tag at the end of the build_image.py script? I usually copy it from there, because I can never remember the Artifact Registry repo path.

nkemnitz:
I guess we could use labels, that would at least allow finding and filtering the image with the correct version.
Maybe the builder could check if scheduler environment and docker image are compatible

supersergiy:
I see your argument, but I feel like "tag" has very specific meaning in docker context, and overloading it to mean just a part of the tag is confusing. How about we rename that argument to "tag_suffix"? 

I've never looked into labels, but it might be a good solution too. Lmk what you prefer

supersergiy:
@nkemnitz made the change. Couldn't verify it works because I'm having an access issue, but hope that will resolve independently

nkemnitz (approved):
Sounds good, no objection.

PR #867, opened 2025-01-25
fix: revert subchunkable to look at 3 chunks, and check for rollover every time

dodamih:


supersergiy (approved):

nkemnitz:
I am OK with merging this to fix the broken blending mode, but adding a test to prevent future regressions would be even better.

PR #868, opened 2025-01-25
fix(docker/pytorch_base): configurable microarch

nkemnitz:
Most of the GCP VM's we use are Skylake without AVX512 support. Lowering the target microarch to v3 resolves the `Illegal instruction (core dumped)` error when importing `graph-tool`.
Pytorch still uses AVX512, if available. They do some runtime checks, I believe.


akhileshh (approved):
Thanks!

PR #871, opened 2025-01-26
feat: install script

supersergiy:
One script that installs `zetta_utils` including all of the necessary apt-get packages and setting all the necessary environment variables. Designed to be usable in the docker file as well to reduce overall duplication.

### DIFF
--- a/install_zutils.py
+++ b/install_zutils.py
@@ -1,0 +1,702 @@
+#!/usr/bin/env python3
+# pylint: disable=wrong-import-position,line-too-long,too-many-branches,too-many-statements
+
+import argparse
+import os
+import platform
+import subprocess
+import sys
+import threading
+from typing import Any
+
+
+def ensure_dependencies():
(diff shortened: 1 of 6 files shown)


install_zutils.py
```diff
@@ -0,0 +1,689 @@
...
+            "libboost-system-dev",
+            "libboost-iostreams-dev",
+            "libjemalloc-dev",
+            "libtbb-dev",
+            "unixodbc-dev",
+        ]
+        run_command("sudo apt-get update", "Updating package lists")
+        apt_install_flags = '-y --no-install-recommends -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"'
+        run_command(
+            f"sudo apt-get install {apt_install_flags} {' '.join(apt_packages)}",
```
nkemnitz, line 643: `sudo` is not available in our current docker image (currently always logged in as `root`)

nkemnitz (approved):
Nice, thanks!

PR #879, opened 2025-01-28
feat: precomp attrs and write_annotations refinements

JoeStrout:
1. Refactored LineAnnotation and AnnotationLayer to make use of `attrs`
2. `AnnotationLayer.write_annotations` now raises a `ValueError` if you supply a bounding box to clear, and any annotations are partially or fully outside that bounds
3. The annotations PUT endpoint silently discards any annotations not within the specified bounds.


supersergiy:
@JoeStrout is this still relevant to be merged or did we evolve away from that?

JoeStrout:
Ugh.  Some of this is still relevant, but will conflict heavily with the refactoring we did (in another branch, now merged) converting AnnotationLayer into a VolumetricBackend.

I'll have to redo the relevant bits by hand, but at this point I should do that after merging all the other outstanding code.

PR #887, opened 2025-02-05
Update cluster creation and install script

nkemnitz:
* ~system node e2-medium -> e2-standard-2, remove manual CPU request overrides~
* keep e2-medium, but enable autoscaling for system nodes (to host the additional `kube-dns` and `konnectivity` pods)
* include helm in zutils install (required for KEDA setup when creating new clusters)


akhileshh:
Looks good, but any reason to explicitly set `--max-pods-per-node "32"` ? The default is 110, what are the benefits of limiting it to 32?

akhileshh (approved):

nkemnitz:
That's just a habit from old days where we had thousands of nodes running and would exhaust the IP address ranges, iirc.

supersergiy (approved):
Would it make sense to add more node pools to these clusters, like we have on `zetta-research`?

PR #888, opened 2025-02-05
adds support for pcg tag to use during installation

akhileshh:


install_zutils.py
```diff
@@ -611,6 +611,9 @@ def main():
     parser.add_argument(
         "--pcg", action="store_true", default=False, help="Whether or not to install PCG"
     )
+    parser.add_argument(
+        "--pcgtag", action="store_true", default="v2.18.3", help="PCG repo tag to use"
```
nkemnitz, line 615: Parser will complain if you specify this as a flag and pass a value
```suggestion
        "--pcgtag", default="v2.18.3", help="PCG repo tag to use"
```

akhileshh: removed

PR #890, opened 2025-02-06
Builder Sequential Fix

supersergiy:


tests/unit/test_builder.py
```diff
@@ -327,3 +329,101 @@ def test_unpickleable_dict():
...
+    with open(path, "w", encoding="utf8") as f:
+        f.write("hello world")
+
+
+def dummy_f(arg):
+    ...
+
+
+@pytest.fixture
+def file_io_funcitons():
```
nkemnitz, line 348:
```suggestion
def file_io_functions():
```

nkemnitz (commented):
Thanks for fixing that one!
I think disallowing parallel build for nontrivial objects broke coverage for the parallel part of a non-parallel build? Or is that code now just unreachable?
https://github.com/ZettaAI/zetta_utils/blob/935d2815fce2d4770728b4c3394cea59b7861b16/zetta_utils/builder/building.py#L154-L161

supersergiy:
Fixed! Indeed, there is no more `parallel_part` in sequential stages, as the parallel part doesn't respect list order.

nkemnitz (approved):

PR #894, opened 2025-02-06
feat: use tinybrain for image downsampling

### LINKED ISSUE #737: 2D downsampling of uint8 data inefficient
All our tensors are passed as NCXYZ to torch and converted to float32. That's not just a copy, but also 4x more memory.

* pytorch interpolate supports 'bilinear' downsampling for uint8 data, but requires NCXY tensor input
* tinybrain has fast AvgPool for (2,2), (2,2,1), (2,2,1,1), as well as for (2,2,2), (2,2,2,1) ndarrays

Another thing to consider is that CloudVolume data already is in Fortran order, which tinybrain expects

```python
data = np.asfortranarray(np.random.randint(0,255, size=(1,1,4096,4096,1), dtype=np.uint8))
```

```python
# Torch CPU, uint8->float32->uint8
%timeit torch.nn.functional.interpolate(torch.from_numpy(data).float(), scale_factor=[0.5,0.5,1.0], mode='trilinear').byte()
84 ms ± 774 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

```python
# Torch GPU, uint8->float32->uint8
%timeit torch.nn.functional.interpolate(torch.from_numpy(data).cuda().float(), scale_factor=[0.5,0.5,1.0], mode='trilinear').byte().cpu()
6.34 ms ± 32.5 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

```python
# Torch CPU, uint8
%timeit torch.nn.functional.interpolate(torch.from_numpy(data).squeeze(-1), scale_factor=[0.5,0.5], mode='bilinear')
162 ms ± 1.02 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

```python
# Torch CUDA, uint8
%timeit torch.nn.functional.interpolate(torch.from_numpy(data).cuda().squeeze(-1), scale_factor=[0.5,0.5], mode='bilinear').cpu()
RuntimeError: "upsample_bilinear2d_out_frame" not implemented for 'Byte'
```

```python
# Tinybrain, uint8
In [16]: %timeit tinybrain.downsample_with_averaging(data.squeeze((0,1)), factor=[2,2])[0]
1.45 ms ± 12.9 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

```python
# Tinybrain, uint8->float32->uint8
%timeit tinybrain.downsample_with_averaging(data.astype(np.float32).squeeze((0,1)), factor=[2,2])[0].astype(np.uint8)
32.1 ms ± 254 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

supersergiy:
fixes #737

### DIFF
--- a/zetta_utils/tensor_ops/common.py
+++ b/zetta_utils/tensor_ops/common.py
@@ -388,7 +388,7 @@ def interpolate(  # pylint: disable=too-many-locals
         allow_slice_rounding=allow_slice_rounding,
     )
 
-    if mode in ("segmentation", "img", "bilinear") and (
+    if mode in ("segmentation", "img", "bilinear", "linear", "trilinear") and (
         scale_factor_tuple is not None
         and (
             tuple(scale_factor_tuple)
--- a/zetta_utils/tensor_ops/common.py
+++ b/zetta_utils/tensor_ops/common.py
@@ -388,7 +388,7 @@ def interpolate(  # pylint: disable=too-many-locals
         allow_slice_rounding=allow_slice_rounding,
     )
 
-    if mode == "segmentation" and (
+    if mode in ("segmentation", "img", "bilinear") and (
         scale_factor_tuple is not None
         and (
             tuple(scale_factor_tuple)
@@ -400,8 +400,10 @@ def interpolate(  # pylint: disable=too-many-locals
         )
         and data.shape[0] == 1
     ):  # use tinybrain
-        result_raw = _interpolate_segmentation_with_tinybrain(
-            data=data, scale_factor_tuple=scale_factor_tuple
+        result_raw = _interpolate_with_tinybrain(
+            data=data,
+            scale_factor_tuple=scale_factor_tuple,
+            is_segmentation=(mode == "segmentation"),
         )
     else:
         result_raw = _interpolate_with_torch(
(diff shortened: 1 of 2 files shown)


zetta_utils/tensor_ops/common.py
```diff
@@ -388,7 +388,7 @@ def interpolate(  # pylint: disable=too-many-locals
         allow_slice_rounding=allow_slice_rounding,
     )
 
-    if mode == "segmentation" and (
+    if mode in ("segmentation", "img", "bilinear") and (
```
nkemnitz, line 391: It's not going to break things, but `tinybrain` has no C implementation for signed integer type averaging (e.g. image encodings). If I read it correctly, it will convert such cases to fp32 and use numpy based interpolation. Probably slower than torch+CUDA

nkemnitz, line 391: Maybe also add `linear` and `trilinear`, just to prevent that the pytorch path is called for 3D volumes, which could be handled by tinybrain. Tinybrain will ignore it and do the downsampling based on the `scale_factor_tuple`.

supersergiy:
cc https://github.com/seung-lab/tinybrain/issues/22

zetta_utils/tensor_ops/common.py
supersergiy: Addressed

supersergiy: Addressed via https://github.com/seung-lab/tinybrain/issues/22

nkemnitz (approved):

Issue #893, opened 2025-02-06
Interpolate doesn't handle separate src and dst correctly when processing multiple resolutions


trivoldus28:
Currently the code just use the [original src](https://github.com/ZettaAI/zetta_utils/blob/d82dacdaa69fd04837f53cb6bce058f1ced7e322/zetta_utils/mazepa_layer_processing/common/interpolate_flow.py#L116) for all resolutions but this is wrong when `dst` is not `src`

trivoldus28:
@supersergiy just FYI since it could affect usability of portal jobs. I think I can work on a fix quickly

supersergiy:
Not sure I understand - what do you mean by "all resolutions"?

trivoldus28:
Let's say you want 8nm & 16nm from 4nm. The flow will try to use the original `src` as source for both 4->8 and 8->16, which is wrong as it should use `dst` as source for 8->16

supersergiy:
Ah gotcha, you're right. Are you on it or should I push a hotfix?

trivoldus28:
Yea I have a working fix that I can push

PR #895, opened 2025-02-06
fix(interpolate): handling different src/dst

trivoldus28:
Fixes https://github.com/ZettaAI/zetta_utils/issues/893


supersergiy (approved):
Thanks for the fix

PR #905, opened 2025-02-14
feat(cv): expose cache_bytes_limit

trivoldus28:
`cache_bytes_limit` was not exposed in `build_cv_layer` like with ts. This is necessary for trainings to not go OOM


dodamih (commented):
I don't think there is a good way around this, but we should note somewhere that if you initially initialise a CV with some cache byte limit (or the default) then trying to reinitialise will not change the byte limit.

trivoldus28:
Where should this note be placed?

trivoldus28:
Actually how does that scenario happen? `_get_cv_cached` is a fn internal to the file and everything that calls it does not change `cache_bytes_limit`, which is only set at construction

trivoldus28:
Ah I see the `_cv_cache` code. If you want I can add `cache_bytes_limit` to the index so you just get different instances

dodamih:
I don't think it's massively better, but I still think just checking if the value is the same is a little safer - we don't want a user who doesn't know the internal workings to reinitialise accidentally thinking that they're just resizing the cache when they're reinitialising and having to download all over again.

trivoldus28:
Maybe the best solution is to check and resize the lru cache to the smallest value seen? Seems easy to do.

Additionally, `cachetools.LRUCache(maxsize=16)` should be customizable. I think my training script uses more than 16. Should be use an environment variable?

dodamih:
If CV lru is resizeable without evicting cache, we should definitely do that. @supersergiy Do you have thoughts on exposing the number of cached CV as an env variable?

trivoldus28:
Seems like a good idea, but maybe more like 1024? Off my head the current EDC spec already uses close to 256 vols:

6 merge thresholds * 4 layers (seg, gt seg, aff or emb, mask) * 9 cutouts = 216

I wonder if there's a drawback to make the LRU _really_ big?

supersergiy:
It sounds like you got a huge training dataset. Are you sure all of this will fit into the workers memory? Because if not, lru cache will not really be helpful here

trivoldus28:
I'm running it with no lru cache (but with disk cache). I guess I don't know whether making a new `CloudVolume()` on every batch would be a significant overhead. Probably not, in which case 256 is reasonable, and probably safer to roll out as a global change.

supersergiy:
Ohh I get it now. I think 1024 is an absolutely reasonable limit. I think until we actually start running OOM because of it we shouldn't worry too much about evicting CloudVolumes from cache. Could shoot for even more than 1024

trivoldus28:
Then I'll go 2048 for future proofing :). I'll also use RRCache instead of LRU so it doesn't update a (potentially very long) list on every access (https://github.com/tkem/cachetools/blob/master/src/cachetools/__init__.py#L210-L214).

We could also think about refactoring it so that `cachetools` tracks sizes across all CVs and prevents accidental OOM altogether. From https://cachetools.readthedocs.io/en/latest/
> In general, a cache’s size is the total size of its item’s values. Therefore, [Cache](https://cachetools.readthedocs.io/en/latest/#cachetools.Cache) provides a getsizeof() method, which returns the size of a given value.

It seems doable. CV provides the `lru.nbytes` interface to query how much it's currently holding ([source](https://github.com/seung-lab/cloud-volume/blob/55238b70023d457aaf016e2814a035ac986e7ccf/cloudvolume/lru.py#L205), note: `self.size` is actually the specified max size).

The only complication of this scheme is that `cachetools` only updates item size on `__setitem__` ([src](https://github.com/tkem/cachetools/blob/master/src/cachetools/__init__.py#L210-L214)), so for accurate tracking you'd need to add extra `_cv_cache[key] = cvol` in `backend.read/write()` after data operations.

trivoldus28:
@supersergiy @dodamih I added lru resize to the cache

dodamih:
@trivoldus28 Sorry to ask, but could you emit a `logger.warning` if the passed in LRU size is bigger than what is in the cache, with a message telling the user that the LRU is not resized? (I still think that this is not very clear relative to just giving the user what they want and letting them handle creation calls with different sizes, but you're the most likely or only user of this feature for the foreseeable future.)

trivoldus28:
Why larger size is not accepted: it's quite often when running a specs that there will be many layer builds referencing the same path, many of which will use the default size. If you allow these to override each other arbitrarily (as often is the case with parallel build or arbitrarily ordered builds) LRU size constraining a layer will most likely be lost and ineffective.

I can add the warning but I think it will be more confusing than not to the users. What do you think, with the above context?

trivoldus28:
Also I'm fairly sure this feature will be useful for training where spawning many dataloaders often cause OOM.

dodamih:
Oh, I have a much better, cleaner idea - only resize if it's explicitly given, and do that every time. Should have thought of this sooner.

supersergiy:
@trivoldus28 @dodamih should we merge this in as is or is more work needed?

trivoldus28:
Yea I need to work on it a little bit more. Should be quick - just a bit distracted.

dodamih:
Shoot, says that it can't be rebased due to conflicts (probably from merging remote). @trivoldus28 Sorry to ask but can you make a new PR that's clean with just the changes? I can do it if you don't care about github history being attributed to you.

dodamih (approved):

dodamih:
Not sure where I'm confused:
image

trivoldus28:
Can you approve it?

trivoldus28:
Yea I'm not exactly sure what rebase and merge is but squash and merge basically does what you suggested by making a clean PR by squashing all changes into a single commit..
image

trivoldus28:
Ah I think what "rebase and commit" is. It probably preserves the commit history of this branch, which is a bad choice in this case because of the intermediary commits that would clutter up the main branch.

PR #925, opened 2025-03-03
AnnotationLayer as a Backend subclass

JoeStrout:
Refactors AnnotationLayer to be a Backend subclass.

### DIFF
--- a/tests/unit/db_annotations/test_precomp_annotations.py
+++ b/tests/unit/db_annotations/test_precomp_annotations.py
@@ -32,7 +32,7 @@ def test_round_trip():
     sf = AnnotationLayer(file_dir, index)
     assert sf.chunk_sizes == [(2000, 2000, 600)]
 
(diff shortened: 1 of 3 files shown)


supersergiy:
This looks solid! Can we make it so that it's implementing a [[Volumetric Backend](https://github.com/ZettaAI/zetta_utils/blob/main/zetta_utils/layer/volumetric/backend.py#L20)] rather than just the generic one?

supersergiy (approved):
Looks great. I can pick it up from there to make a layer, and then we can push it to @dodamih to make sure subchunkable works

supersergiy:
Something seems to be up with this PR. Can you confirm that tests don't hang locally? They don't hang on other PRs I just re-ran

JoeStrout:
Yeah, it appears that tests/unit/mazepa/test_id_generation.py is hanging, even locally.  I'll see if I can figure out why.

PR #927, opened 2025-03-06
some improvements to synapse sv lookup scripts

akhileshh:


JoeStrout (approved):

supersergiy:
@akhileshh the problem here is that pylint actually crashes because of supposedly illegal syntax here. perhaps fixing the mypy issues would be useful. mypy itself wasn't disabled for scripts, it's actually enabled. the reasoning here is that -- every type buy is actually a bug, users don't have to add types to `scripts` but if there's a bug then it shouldn't be merged

akhileshh:
Alex's csv should include supervoxels going forward so closing this.

PR #928, opened 2025-03-10
fix: enable idle_timeout only for scaledjobs

akhileshh:


supersergiy (approved):
Looks good, thanks for adding it back

PR #930, opened 2025-03-11
fix(bbox_strider): ensure offset aligned with voxel resolution

nkemnitz:


nkemnitz:
Closed in favor of #931

PR #932, opened 2025-03-12
feat: graphene layer support

supersergiy:


zetta_utils/common/signal_handlers.py
```diff
@@ -4,18 +4,13 @@
 
 @contextmanager
 def custom_signal_handler_ctx(fn, target_signal):  # pragma: no cover
-    # save original SIGINT handler
     original_sigint_handler = signal.getsignal(target_signal)  # eg signal.SIGINT
```
dodamih, line 7: Needs to be renamed to `original_signal_handler`.

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
```diff
@@ -637,7 +637,7 @@ def _path_join_if_not_none(base: str | None, suffix: str) -> str | None:
 
 def _expand_bbox_resolution(  # pylint: disable=line-too-long
     bbox: BBox3D,
-    dst_resolution: Vec3D,
+    dst_resolution: Vec3D[float],
```
dodamih, line 640: Was this causing typing issues? Or just adding more strict typing?

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
```diff
@@ -829,3 +829,5 @@ def flow(  # pylint:disable=too-many-branches, too-many-statements
             delete_if_local(*dst_temps)
         if self.clear_cache_on_return:
             clear_cache(*op_args, **op_kwargs)
+
+        return tasks
```
dodamih, line 833: Does it make sense to return tasks rather than the task's output?

dodamih (changes requested):

zetta_utils/mazepa_layer_processing/common/volumetric_apply_flow.py
supersergiy: Task output is not known at this time

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
supersergiy: No issues with mypy, I think vscode has some issues that I was trying to fix. I don't think this change is necessary. Do you want me to remove it?

zetta_utils/common/signal_handlers.py
supersergiy: k

zetta_utils/mazepa_layer_processing/common/subchunkable_apply_flow.py
dodamih: Yeah - I think looser is better here.

PR #943, opened 2025-03-27
fix(simple_inference_runner): fix GPU memleak

nkemnitz:


supersergiy (approved):
Looks good, thanks for the fix!

PR #945, opened 2025-03-28
feat: cave_secret.json for remote workers

dodamih:


akhileshh (approved):
Looks good, thanks! Just wondering if this will ever be required in training?

PR #968, opened 2025-05-20
feat: O(n) scanline algorithm for subchunkable + chore: deprecate dst_tighten_bounds

dodamih:


supersergiy:
@dodamih is there a reason to update the `black` version?

PR #987, opened 2025-07-11
feat: use install_zutils.py everywhere + pcg support

dodamih:


.github/workflows/docs_deployment.yaml
```diff
@@ -41,13 +41,14 @@ jobs:
           python-version: 3.11
           architecture: x64
           cache: 'pip'
-
       - name: Install libboost
         run: sudo apt install libboost-dev
```
supersergiy, line 45: Is libboost not installed by `install_zutils`?

dodamih: Running skip_apt since everything else exists in the docker image.

PR #1003, opened 2025-07-30
Add builder for VolumetricIndex ctor, and bbox/res option for build_annotation_layer.

JoeStrout:


zetta_utils/layer/volumetric/index.py
```diff
@@ -331,4 +338,5 @@ def line_intersects(
         return self.bbox.line_intersects(pointA, pointB, self.resolution)
 
 
+builder.register("VolumetricIndex")(VolumetricIndex)
```
supersergiy, line 341: This can be done as a class decorator

PR #1012, opened 2025-08-11
Fix JIT trace/export training corruption

nkemnitz:
Passing the model to the JIT export subprocess will still share CUDA tensors. The trace can silently fail ([see Warning](https://docs.pytorch.org/docs/stable/generated/torch.jit.trace.html)) and modify the behavior of the model during training.

This PR:
* Allows users to independently disable JIT and ONNX export
* Fixes a few warnings related to (unnecessary) branching in our crop method (but that didn't solve the issues I encountered)
* completely recreates the model inside the JIT export subprocess to avoid any side effects on the main process model - so slightly larger (GPU) memory footprint
* still keeps the spawned process to avoid memory leaks
* also fixes a bug (I think it is) in which our exports were done in training mode


nkemnitz:
I have no strong opinion on the defaults. For this PR I just replicated the current behavior

nkemnitz:
OK, after updating drivers and being forced to reinstall cudatoolkit, I can no longer replicate the exploding loss caused by JIT trace and/or export... so I reverted the deepcopy of the model including CUDA tensors. But at least added a bunch of tests to ensure checkpoint methods are correctly called / not called.

trivoldus28:
The model state was changing unexpectedly?

tests/unit/training/lightning/test_default_trainer.py
```diff
@@ -9,3 +21,257 @@ def test_default_trainer():
...
+        "lightning_module": mocker.patch.object(
+            ZettaDefaultTrainer, "lightning_module", new_callable=PropertyMock
+        ),
+    }
+
+
+def test_save_checkpoint_calls_exports_when_enabled(
+    trainer_mocks, mock_model, mock_trace_input, mock_lightning_module
+):
+    """Test that save_checkpoint calls export functions when exports are enabled."""
```
trivoldus28, line 74: is this test necessary? isn't it simpler just to add an assert line in the code that checks at least 1 export function is enabled?

tests/unit/training/lightning/test_default_trainer.py
```diff
@@ -9,3 +21,257 @@ def test_default_trainer():
...
+        trainer_mocks["jit_export"].assert_called_once_with(
+            mock_model, (mock_trace_input,), filepath, "test_model"
+        )
+        trainer_mocks["super_save"].assert_called_once()
+
+
+def test_save_checkpoint_skips_jit_when_disabled(
+    trainer_mocks, mock_model, mock_trace_input, mock_lightning_module
+):
+    """Test that save_checkpoint skips JIT export when disabled."""
```
trivoldus28, line 107: seems not necessary. if jit.trace is disabled in the input param it's clear that it's disabled.

tests/unit/training/lightning/test_default_trainer.py
```diff
@@ -9,3 +21,257 @@ def test_default_trainer():
...
+        trainer.save_checkpoint(filepath)
+
+        trainer_mocks["jit_export"].assert_called_once_with(
+            mock_model, (mock_trace_input,), filepath, "test_model"
+        )
+        trainer_mocks["onnx_export"].assert_not_called()
+
+
+def test_save_checkpoint_skips_exports_non_global_zero(trainer_mocks):
+    """Test that save_checkpoint skips exports when not global zero rank."""
```
trivoldus28, line 165: this seems useful, but would miss additional exports in the future

tests/unit/training/lightning/test_default_trainer.py
trivoldus28: > That's different. I am OK if the user disables both exports, so that assert would be too restricting, anyway.
Right, I misread the test.

But yea, still seems trivial? Is it really doing much more than a simple line coverage test? (i.e., call the save function & check that the saved file exists, etc.). Making sure the functions are called seems very marginally better.

trivoldus28:
Have you actually confirmed that the training state changing is the root cause of the issue though? Seems unlikely to me that some external libraries/drivers would be specific enough to torch to change that specific variable and cause a corruption.

I would actually guess that on a configuration with exploding losses, the test would pass and you'd still get that error.

PR #1018, opened 2025-08-15
fix(trainer): enable_progress_bar=False bypasses RichProgressBar

nkemnitz:
`enable_progress_bar=False` otherwise fails with
``MisconfigurationException: Trainer was configured with `enable_progress_bar=False` but found `RichProgressBar` in callbacks list.``


trivoldus28 (approved):
lgtm, thanks!

PR #1045, opened 2025-09-25
feat: performance / semaphore tracking, worker log suppression, log improvement

dodamih:


zetta_utils/geometry/bbox_strider.py
```diff
@@ -240,7 +240,7 @@ def shape(self) -> Vec3D[int]:  # pragma: no cover
 
     def get_all_chunk_bboxes(self) -> List[BBox3D]:
         """Get all of the chunks."""
-        if self.num_chunks > multiprocessing.cpu_count():
+        if self.num_chunks > 100 * multiprocessing.cpu_count():
```
trivoldus28, line 243: The `100*cpu_count` threshold doesn't really make sense - you're just increasing the threshold (penalizing?) larger head nodes with more CPUs than smaller ones.

I think it makes more sense as a threshold that does not depend on machine's cpu count, so:
```
TASK_THRESHOLD = 4*128  // pick a reasonable number that would be slow for a single thread to execute
...
if self.num_chunks > TASK_THRESHOLD:
   with multiprocessing.Pool() as pool_obj:
...
```

dodamih: Reasonable - the overhead should be mostly constant.

trivoldus28: Ah if you worry about overhead, you can limit the number of processes spawned, based on chunk count:
```
n_procs = self.num_chunks // 128  # 128 chunks per proc, or however much
n_procs = min(multiprocessing.cpu_count(), n_procs)  # don't spawn more than cpu count
with multiprocessing.Pool(n_procs) as pool_obj:
...
```

tests/unit/mazepa/test_id_generation.py
```diff
@@ -249,12 +253,52 @@ def _gen_id_calls(_) -> dict[str, str]:
...
-    with ctx.Pool(processes=2) as pool:
-        result = pool.map(_gen_id_calls, range(2))
-
-    assert result[0] == result[1]
+    # Create two separate processes with different PYTHONHASHSEED values
+    # to ensure ID generation is consistent across different hash seeds.
+    # This workaround is necessary due to fork being the default
+    # multiprocessing method, meaning we cannot use 'spawn' method to test.
+
+    project_root = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
```
nkemnitz, line 261: ?!

tests/unit/mazepa/test_id_generation.py
```diff
@@ -249,12 +253,52 @@ def _gen_id_calls(_) -> dict[str, str]:
...
 def test_persistence_across_sessions() -> None:
-    # Create two separate processes - spawn ensures a new PYTHONHASHSEED is used
-    ctx = multiprocessing.get_context("spawn")
-    with ctx.Pool(processes=2) as pool:
-        result = pool.map(_gen_id_calls, range(2))
-
-    assert result[0] == result[1]
+    # Create two separate processes with different PYTHONHASHSEED values
+    # to ensure ID generation is consistent across different hash seeds.
+    # This workaround is necessary due to fork being the default
```
nkemnitz, line 258: I don't understand. Why can't we spawn a new process in the test anymore? This test should not depend on how the workers are created. It's a test to ensure IDs stay the same across different invocations of `zutils run`

dodamih: So the issue is that you cannot mix `spawn` and `fork` - once you've used `fork` as your method, you cannot `spawn` any processes anymore, and PYTHONHASHSEED is same across forked processes, meaning that we aren't testing IDs across different invocations anymore.

tests/unit/mazepa/test_id_generation.py
dodamih: Seemed like a reasonable way to get the root location relative to the test file no matter where you run it from.

tests/unit/mazepa/test_id_generation.py
dodamih: You should be able to normally, but there's some weird interaction with pytest that just makes it hang forever.

trivoldus28: None of the chatbots think that "mp.set_start_method('fork') breaks mp.get_context('spawn')" though...

tests/unit/mazepa/test_id_generation.py
dodamih: test_id_generation.py -> mazepa -> unit -> tests -> zetta_utils. I guess I can do `zetta_utils.__path__` instead.

tests/unit/mazepa/test_id_generation.py
dodamih: It also works fine on a minimal example even in pytest, so it's something more insidious.

dodamih: Minimal example - try this:

```
import multiprocessing as mp
import pytest

mp.set_start_method('fork')

def simple_task(x):
    return x * 2

def test_multiprocessing_context():

    ctx = mp.get_context("spawn")
    with ctx.Pool(processes=2) as pool:
        results = pool.map(simple_task, [1, 2, 3, 4])
        assert results == [2, 4, 6, 8]

test_multiprocessing_context()
```

Barfs with 
``` File "/home/dodam/miniconda3/envs/zetta_new/lib/python3.12/multiprocessing/spawn.py", line 131, in _main
    prepare(preparation_data)
  File "/home/dodam/miniconda3/envs/zetta_new/lib/python3.12/multiprocessing/spawn.py", line 246, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "/home/dodam/miniconda3/envs/zetta_new/lib/python3.12/multiprocessing/spawn.py", line 297, in _fixup_main_from_path
    main_content = runpy.run_path(main_path,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen runpy>", line 287, in run_path
  File "<frozen runpy>", line 98, in _run_module_code
  File "<frozen runpy>", line 88, in _run_code
  File "/home/dodam/zetta-packages/zetta_utils/test_mp_context.py", line 4, in <module>
    mp.set_start_method('fork')
  File "/home/dodam/miniconda3/envs/zetta_new/lib/python3.12/multiprocessing/context.py", line 247, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
```
but this is an infinite loop, which is why pytest just hangs forever.

dodamih: Adding separate context to every mp call should fix it, so I'll do that.

trivoldus28: The error is `mp` complaining `set_start_method` being called multiple times for some reason

dodamih: Yeah, but can't do that easily just for pytest. Separate context is probably easier...

nkemnitz: Yeah, seems the spawned processes already have their default start method set and attempt to run the top-level `mp.set_start_method`, which is not allowed. The error disappears if you place `mp.set_start_method('fork')` inside a `if __name__ == "__main__":` guard (so the spawned workers skip that call), but that's also not what we want.

Creating a separate context for every mp call seems like a more readable approach, anyway, so I am fine with that.

PR #1049, opened 2025-10-02
chore: update some seunglab dependencies

dodamih:
CI didn't run tests because no python file changes, but I ran unit tests manually and they passed.


trivoldus28 (approved):
Thanks!

PR #1062, opened 2025-10-15
feat(imgaugs): add specifiable `targets`

trivoldus28:
These are also necessary for the CI training example


zetta_utils/augmentations/imgaug.py
```diff
@@ -114,13 +117,21 @@ def _ungroup_kwargs(
 @builder.register("imgaug_readproc")
 def imgaug_readproc(
     *args,  # the zetta_utils builder puts the layer/layerset as the first argument
+    targets: Iterable[str] | None = None,
```
dodamih: Sorry to ask, but could you add a comment to this please? Or (better) use something like

```
    """
    Read processor using imgaug.

    :param *args:
    :param ...
    """
```
Also not a fan of `*args` + `assert` instead of just using `data` (which makes it a little clearer that it conforms to the proc protocol), but that's fine.

trivoldus28:
@dodamih @supersergiy Changed nothing but comments but now I'm getting this error. Do you know what's going on?
image

zetta_utils/augmentations/imgaug.py
```diff
@@ -13,10 +15,11 @@
 from zetta_utils.tensor_ops import common, convert, crop_center
 from zetta_utils.tensor_typing import Tensor, TensorTypeVar
 
-SizedTypeVar = TypeVar("SizedTypeVar", bound=Sized)
+# SizedTypeVar = TypeVar("SizedTypeVar", bound=Sized)
```
nkemnitz: Same here

PR #1063, opened 2025-10-15
feat(train): adds pod affinity for training

akhileshh:


trivoldus28 (approved):
`required_zones` seems to work - thanks!

PR #1071, opened 2025-10-27
hotfix: change start method to spawn

dodamih:


zetta_utils/__init__.py
```diff
@@ -3,12 +3,16 @@
...
 import warnings
+import multiprocessing
 
 from .log import get_logger
 
 # Set global multiprocessing threshold
 MULTIPROCESSING_NUM_TASKS_THRESHOLD = 128
 
+# Unnecessary after 3.14 - not encapsulated in main due to workers
+multiprocessing.set_start_method("forkserver", force=True)
```
trivoldus28: Not an expert but all of the best practice guides I've come across recommend to never put this unguarded.

zetta_utils/__init__.py
```diff
@@ -3,12 +3,17 @@
...
 import sys
 import warnings
+import multiprocessing
 
 from .log import get_logger
 
 # Set global multiprocessing threshold
 MULTIPROCESSING_NUM_TASKS_THRESHOLD = 128
 
+# Set start method to `forkserver`` if not set elsewhere
```
trivoldus28: Add a comment here saying that this is done because `fork` is not compatible with dependencies, and that this code will be irrelevant by python3.14 when forkserver is made the default?

zetta_utils/mazepa_addons/configurations/worker_pool.py
```diff
@@ -78,15 +79,14 @@ def setup_local_worker_pool(
 ):
     """
     Context manager for creating task/outcome queues, alongside a persistent pool of workers.
+    Note that worker pools will inherit the current process' multiprocessing start method.
     """
     with monitor_resources(resource_monitor_interval):
         pool = pebble.ProcessPool(
             max_workers=num_procs,
-            context=multiprocessing.get_context("fork"),
+            context=multiprocessing.get_context("spawn"),
```
trivoldus28: Add comment that `fork` causes issues with cloudvolume sharded reads

PR #1091, opened 2025-11-26
fix: replace all remaining forks with forkserver

dodamih:


dodamih:
@trivoldus28 Haven't, but forkserver isn't that much slower than fork (and is often faster). The first time can be a little slow, but that's once per run and happens during init. This change was necessary because forking caused gRPC barfs - having anything that used Google's OAuth initialised and then forking caused gRPC to show warning messages / wasn't good practice in general for thread safety.

There's still some issues (see #1092) leftover, but that can be solved by switching from spawn to forkserver as well for the local pool and not force-initialising the head node's spawn method if it is forkserver.

dodamih:
The other reason for this change was that forking after going inside the `graceful_shutdown` context was resulting in confusing error messages - when the forked processes exited, it resulted in a SIGTERM, which probablistically (depending on the timing) raised an error. (Either "Detected signal" or "Process ForkPoolWorker... RuntimeError: Cancelled" message.) There's other ways to fix this, but the cleanest way was to use forkserver and fork from before entering the context.

trivoldus28:
Please check performance. I'm concerned all new processes reinitialize (ie
reimporting np, torch, etc)

dodamih:
This is the warning message that I'm concerned about btw: 
image

PR #1215, opened 2026-04-08
feat(reduce_operations): bump fn

nkemnitz:
Got the same results as with chunkflow, so I am reasonably confident. But I decided against their normalization approach - I don't see the floating point problems (yet), and I want to treat the pixel centers as corners for blending, similar to how we do it in our other blending modes.


dodamih (approved):
Thanks Nico!

PR #1217, opened 2026-04-11
resources and semaphore tracking in run-info; other improvements

akhileshh:
Extends the GCS tracker sidecar to also publish **semaphore timing** and **system resource** stats per pod, then aggregates everything on the head node into a unified view in `RUN_DB[run_id]`. One Firestore document per pod with all three stat types.

- **Decoupled the sidecar from mitmproxy.** New `StatsCollector` abstraction; the push thread runs unconditionally, GCS collector is registered only when mitmproxy is available.
- **Semaphore collector.** Sidecar reads `/dev/shm/zetta_semaphore_timing_*` written by the main container's `TimingTracker`.
- **Resource collector.** Main container runs a new `write_resource_stats_file()` context manager alongside `monitor_resources()` that writes `ResourceMonitor.get_summary_stats()` to `/tmp/resource_monitor_stats.json`; sidecar reads it. `ResourceMonitor` itself untouched. Gets accurate main-container psutil/pynvml numbers.
- **Unified pod-stats collection.** `gcs-stats-proxy` → `pod-stats`. `WORKER_TYPE` env var (set by the worker pool spec) is stamped on every doc for per-worker-type aggregation.
- **Unified head-node aggregator.** `aggregate_pod_stats(run_id)` does **one Firestore read + one write**, dispatching to gcs/semaphore/resource sub-aggregators. Promotes `total_egress_gib` to a top-level field for query-friendliness. Live cost log enriched with bottleneck semaphore + fleet CPU/memory; cached in-process so the log doesn't add a Firestore read.

#### Instrumentation must never crash the run

- Wrapped `_update_costs` and the final `update_run_info` in `run_ctx_manager.finally`. Replaced asserts in `compute_costs` with warnings.
- Heartbeat is intentionally **not** wrapped: a stale-heartbeat GC tears down k8s resources, so swallowing the error would let the run keep doing work it's about to lose.
- Added `timeout` kwarg to `DBLayer.query` / `FirestoreBackend.query`. `compute_costs` passes `timeout=30s` so a transient network blip can't block the RepeatTimer thread for the SDK's default 5-minute deadline.
- `_cleanup_pod_stats` now deletes all per-pod docs in one Firestore RPC via `bulk_writer`, instead of looping.

#### Cleanup

- **Region check rework.** Explicit positive lists for US/EU/ASIA, all 6 predefined dual-regions, `bucket.data_locations` passthrough for configurable dual-regions, documented exclusion set (London/Zurich for EU, HK/Jakarta for ASIA). Conservative: unknown regions return `False` so egress is never silently missed. Fixes the false-positive warning observed for NAM4 buckets.
- Removed dead `extract_bucket_from_path`, `check_bucket_region_alignment`, `extract_gcs_paths_from_spec`, `validate_region_alignment_for_run`.
- `_get_group_taskqueue_and_contexts` no longer unpacks 8–11 group fields inline; two private helpers absorb the noise. `k8s.pod` / `k8s.deployment` stay generic.
- Dropped legacy `GCS_STATS_DB` and `aggregate_gcs_stats`.


zetta_utils/cloud_management/resource_allocation/gcloud/gcs.py
```diff
@@ -16,30 +16,108 @@
...
 logger = log.get_logger("zetta_utils")
 
-# GCS region to GCE region mapping
-# GCS uses location names that may differ from GCE region names
-# Multi-region locations and their corresponding regions
-MULTI_REGION_MAPPING = {
-    "US": ["us-east1", "us-east4", "us-west1", "us-west2", "us-west3", "us-west4", "us-central1"],
-    "EU": ["europe-west1", "europe-west2", "europe-west3", "europe-west4", "europe-north1"],
-    "ASIA": ["asia-east1", "asia-east2", "asia-northeast1", "asia-south1", "asia-southeast1"],
+# Goal: detect cross-region egress no matter what. The function returns True
```
dodamih: This docstring seems orphaned? I'm guessing that this is for `is_region_compatilbe`

zetta_utils/cloud_management/resource_allocation/k8s/gcs_tracker.py
```diff
@@ -310,34 +321,75 @@ def run_proxy(run_id: str):
...
     logger.info(f"Starting mitmproxy on port {PROXY_PORT}")
-    process = subprocess.Popen(  # pylint: disable=consider-using-with
+    return subprocess.Popen(  # pylint: disable=consider-using-with
         mitm_cmd,
         stdout=sys.stdout,
         stderr=sys.stderr,
     )
 
+
+def run_proxy(run_id: str):
```
dodamih: This needs to be renamed - it's not just running a proxy anymore, right? So it should probably be `run_stats_sidecar` or something like that.

zetta_utils/cloud_management/resource_allocation/k8s/gcs_tracker.py
```diff
@@ -310,34 +321,75 @@ def run_proxy(run_id: str):
...
+    """
+    Run the pod stats sidecar.
+
+    Always starts the stats push loop. Optionally starts mitmproxy for
+    GCS request interception when mitmproxy is available.
```
dodamih: Are there situations where we won't have mitmproxy?

zetta_utils/run/costs.py
```diff
@@ -55,7 +55,12 @@ def _get_start_time(node_info: DBRowDataT) -> float:
                 lname, build_firestore_layer(lname, DATABASE_NAME, project=PROJECT)
             )
             skus = layer.query({"-regions": [node_region], "class": [machine_class]}, union=False)
-            assert len(skus) == 1, (lname, node_region, machine_class, skus)
+            if len(skus) != 1:
```
dodamih, line 72: Curious - when can this happen?

dodamih (approved):

PR #1218, opened 2026-04-12
refactor: replace pebble process timeout with signal.alarm

dodamih:
Use SIGALRM in the worker process instead of spawning a separate process for task timeout enforcement. This eliminates one process creation per timed task execution.

### DIFF
--- a/zetta_utils/mazepa/tasks.py
+++ b/zetta_utils/mazepa/tasks.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
 import functools
+import signal
 import sys
 import time
 import traceback
 import uuid
-from concurrent.futures import TimeoutError as PebbleTimeoutError
 from typing import (
     Any,
     Callable,
@@ -20,7 +20,6 @@ from typing import (
 )
 
 import attrs
-from pebble import concurrent
 from typing_extensions import ParamSpec
 
 from zetta_utils import log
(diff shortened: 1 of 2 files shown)


nkemnitz (commented):
If the task is stuck in a C extension call or similar and we want to interrupt that, there is no way around the separate process. I think you had something similar happen recently with waterz?

If this caveat is acceptable, I would like to include a comment briefly explaining the decision for posterity

zetta_utils/mazepa/tasks.py
```diff
@@ -75,13 +74,17 @@ def _call_task_fn(self, debug: bool = True) -> R_co:
...
         else:
-            future = concurrent.process(timeout=self.runtime_limit_sec)(self.fn)(
-                *self.args, **self.kwargs
-            )
+
+            def _timeout_handler(signum, frame):
+                raise exceptions.MazepaTimeoutError(f"Task '{self.id_}' took too long.")
+
+            old_handler = signal.signal(signal.SIGALRM, _timeout_handler)
+            signal.alarm(int(self.runtime_limit_sec))
```
nkemnitz, line 82: `runtime_limit_sec` is currently typed as `float`. If we go for signal-based interrupt, we should enforce `int` there, otherwise 0.9 gets silently truncated to 0.

dodamih: Good point - thanks, will fix.

dodamih:
@nkemnitz Hmm, also good point. I'll close this for now since I think slowness in spawn is acceptable.

PR #1222, opened 2026-04-16
feat: Refactor run_worker / resource monitoring + Better signal handling

dodamih:
## Overview

Collapse `mazepa.run_worker` into `mazepa.run_worker_manager` so all k8s worker pods go through one entry point, and fix the signal-handling, orphan-cleanup, and resource-monitoring issues that became visible (or newly relevant) once that collapse landed.

Five commits, intended to be read in order.

---

### 1. `88767e9d` — refactor: collapse run_worker into run_worker_manager path

Make `run_worker` an internal helper and route **all** k8s worker pods through `mazepa.run_worker_manager` regardless of `num_procs` / `semaphores_spec`. Centralize `PoolActivityTracker` and resource monitoring ownership in the pool manager so worker processes become pure task executors that no longer manage their own lifecycle state or do duplicate pod-wide measurement.

- `get_mazepa_worker_command` always emits `mazepa.run_worker_manager`; the `num_procs==1 && semaphores_spec is None` branch that previously emitted `mazepa.run_worker` is gone.
- Drop `@builder.register("run_worker")` on `mazepa/worker.py` and the `mazepa.run_worker` re-registration in `mazepa_layer_processing/__init__.py`. `run_worker` stays as a plain function for internal use.
- `run_worker` signature: drop `pool_name` and `resource_monitor_interval`; accept `activity_tracker: PoolActivityTracker | None` instead. Remove the `owns_activity_tracker` / `create_shared_memory` / `unlink` block — `setup_local_worker_pool` is the sole creator and unlinker of the shm.
- `run_local_worker` takes the tracker instance and threads it through. `setup_local_worker_pool` passes its already-constructed tracker via pebble's schedule args (picklable because `PoolActivityTracker` is `attrs.frozen` around a single `str`).
- Merge `write_resource_stats_file` into `monitor_resources`. One `ResourceMonitor`, one `RepeatTimer`, one psutil sample per interval, replacing the two parallel monitors that each sampled the same pod-wide counters. `monitor_resources` now takes an optional `stats_file_path` (default `/tmp/resource_monitor_stats.json`). Errors in the tick are caught so file-write failures don't silently kill the timer thread. Delete `zetta_utils/common/resource_stats_file.py` and its test.
- Move resource monitoring out of `run_worker` entirely — it now lives only in `setup_local_worker_pool`, which means the pool's manager process is the single writer to the stats file. **Fixes a pre-existing regression**: before this change, multi-proc k8s pods never wrote `/tmp/resource_monitor_stats.json` because `run_local_worker` never passed `resource_monitor_interval` through to `run_worker`, so the sidecar's `resource_stats` collector saw `None` on every cycle for `num_procs > 1` pods.
- Tests: `test_worker.py` fixtures yield the tracker instance directly and tests pass `activity_tracker=tracker` instead of `pool_name=pool_name`. Deleted `test_worker_creates_own_activity_tracker` since workers no longer create their own tracker. `test_end_to_end_workflow_sqs.py` idle-worker fixture updated to match.

---

### 2. `f1965446` — fix: unregister attached pool activity shm from resource_tracker

Same fix commit `85717048` applied to `TimingTracker`, now applied to `PoolActivityTracker`. When attaching processes call `shared_memory.SharedMemory(name=...)` without `create=True`, Python's `resource_tracker` still registers the segment and tries to unlink it on process exit even though the process didn't create it ([bpo-38119](https://bugs.python.org/issue38119)). In a pool, that means the first worker to exit unlinks the segment out from under the other workers and the manager's explicit `unlink()`.

Workaround: after attaching, call `resource_tracker.unregister(f"/{name}", "shared_memory")` so only the head node (which called `create_shared_memory`) owns cleanup. The manager's `finally` block still calls `unlink()` explicitly, which is now the sole source of truth for segment lifetime.

Two-line change to match the pattern already in `semaphores.py:47`.

---

### 3. `d10c2c00` — feat: graceful signal handling in mazepa worker pools

Make Ctrl-C in `execute_locally` / `local_test=True` runs actually respect `require_interrupt_confirm`, and make k8s SIGTERM on worker pods produce a graceful exit that acks the in-flight task before dying. Replace the old blanket `reset_signal_handlers()` (which set SIGINT/SIGTERM to `SIG_DFL` — kernel terminate — and silently killed pebble children before the head's confirm prompt could run) with a mazepa-specific signal setup that ignores terminal Ctrl-C in worker children and converts SIGTERM/SIGHUP into a flag the pull loop consults between batches.

- **`worker_init`** (`mazepa/worker.py`): drop `reset_signal_handlers()` call. Add module-level `_shutdown_event` (`threading.Event`, per-process under spawn) and `_graceful_exit` handler that flips it. New `_install_worker_signal_handlers()` helper: `SIGINT=SIG_IGN` (parent owns the confirm decision), `SIGTERM`/`SIGHUP = _graceful_exit`. `_check_exit_conditions` consults `_shutdown_event` first and returns `("sigterm", True)`; the pull loop already re-checks conditions between batches, so the current task completes and gets acked before exit.
- **`setup_local_worker_pool`**: `finally` block now forwards SIGTERM via `psutil.Process(os.getpid()).children(recursive=False)` before `pool.join()`. Without forwarding, `pool.join()` would wedge because `pool.stop()` doesn't interrupt running tasks and `run_worker`'s pull loop is infinite. `recursive=False` avoids double-signaling SQS upkeep handler subprocesses, which are torn down by each worker's own cleanup path.
- **`run_worker_manager`**: installs its own SIGTERM/SIGHUP handler that flips a local `threading.Event`. Polling loop checks the flag at the top of each iteration and breaks on set. ExitStack unwind then triggers `setup_local_worker_pool`'s forwarder — single source of truth for SIGTERM delivery to pool children.
- **`common/signal_handlers.py`**: remove now-unused `reset_signal_handlers`. Add `configure_pool_signals` initializer for short-lived `multiprocessing.Pool` children (used by `bbox_strider.get_all_chunk_bboxes` and `volumetric_apply_flow._create_tasks`): SIGINT=SIG_IGN (parent owns the confirm decision), SIGTERM/SIGHUP=SIG_DFL so `pool.terminate()` and explicit `kill` work. Named to avoid collision with mazepa's `worker_init` and to be accurate about what it configures.
- Fork-pool call sites switched from `reset_signal_handlers` to `configure_pool_signals`. Before, these pools set SIGINT to `SIG_DFL`, so Ctrl-C during a `pool.map` silently killed the fork children and broke the parent's confirm prompt. After, fork children ignore SIGINT (parent's confirm decision wins) and still die cleanly via `pool.terminate()` or kernel SIGTERM.

**Behavior on Ctrl-C in `local_test`**: parent prompts, pebble mazepa workers ignore SIGINT and keep pulling tasks. Decline → workers keep running. Confirm → `KeyboardInterrupt` unwinds ExitStack which forwards SIGTERM to pool children which gracefully exit after acking.

**Behavior on k8s SIGTERM**: manager flag flips, polling loop breaks, ExitStack unwinds, forwarder hits pebble workers, workers finish current task + ack + exit inside the grace period.

Tests: `test_worker_init_*` now mock `_install_worker_signal_handlers` instead of `reset_signal_handlers`. Same assertion pattern.

---

### 4. `33b79c49` — refactor: centralize idle_timeout decision in run_worker_manager

Move the pool-wide idle-timeout check out of `run_worker` and into `run_worker_manager`'s polling loop. Before, every pebble worker read the shared memory tracker each iteration to decide "have we all been idle long enough?" — N workers, N redundant reads, N racing exits. Now the decision is made once per pool, in the one place that already coordinates worker lifecycle.

- `worker.py`: `_check_exit_conditions` drops `idle_timeout` and `activity_tracker` params; only checks `_shutdown_event` and `max_runtime`. `_handle_idle_state` drops `idle_timeout` param; diagnostic log still shows last-activity duration and active-worker count. `run_worker` drops its `idle_timeout` param entirely. Workers still **write** to `activity_tracker` via `_process_task_batch`'s `increment/decrement/update_activity_time` — that's how the manager observes per-worker task state across the process boundary.
- `worker_pool.py`: `run_local_worker` drops `idle_timeout` param. `setup_local_worker_pool` still creates the `PoolActivityTracker` shm when `idle_timeout is not None`, still passes the tracker to workers for write-side updates, but stops passing `idle_timeout` to workers. `run_worker_manager` constructs its own local `PoolActivityTracker` using the same pool_name derivation (`f"{task_queue.name}_{outcome_queue.name}"`). Both references address the same shared memory by name. Polling loop checks `activity_tracker.check_idle_timeout` at the top of each iteration alongside the SIGTERM flag and breaks out with "Pool idle timeout exceeded" when triggered. ExitStack unwind then forwards SIGTERM to pool children — same unified shutdown path as k8s eviction.
- New `tests/unit/mazepa_addons/test_worker_pool.py` with direct tests for `run_worker_manager`:
  - `test_run_worker_manager_exits_on_idle_timeout` — creates the expected shm by name, runs the manager with a short `idle_timeout`, asserts it exits inside the time budget. Catches pool-name drift between `setup_local_worker_pool` and `run_worker_manager`, and deleted-idle-check regressions.
  - `test_run_worker_manager_exits_on_sigterm` — sends SIGTERM from a helper thread while the manager polls, asserts it exits via the shutdown-flag path. Catches missing SIGTERM handler install and broken flag checks.
  - `restore_signal_handlers` autouse fixture saves/restores SIGTERM and SIGHUP around each test so the manager's `signal.signal` calls don't leak to subsequent tests.
- Deleted `test_worker_with_activity_tracker_idle_timeout` and `test_worker_with_activity_tracker_no_idle_when_tasks_processing` from `test_worker.py` (behavior moved out of `run_worker`; `test_pool_activity.py` still covers `check_idle_timeout`'s decision logic directly). Deleted `queues_with_idle_worker` fixture and `test_idling_worker` from `test_end_to_end_workflow_sqs.py`.

---

### 5. `3d4618f8` — feat: PR_SET_PDEATHSIG=SIGTERM in worker_init

If `run_worker_manager` dies via SIGKILL, OOM, or any path that bypasses `setup_local_worker_pool`'s finally-block forwarding, pebble pool workers are orphaned and would linger until they notice a broken queue or similar side effect. Ask the kernel to deliver SIGTERM to them on parent death via `prctl(PR_SET_PDEATHSIG, SIGTERM)`. Composes with the graceful SIGTERM handler from commit 3: orphaned workers auto-receive SIGTERM, flip `_shutdown_event`, finish the current task, ack, and exit cleanly instead of hanging.

- New `_set_pdeathsig_sigterm()` helper in `mazepa/worker.py` uses `ctypes.CDLL(None)` to call libc's `prctl`. Wrapped in `try/except` so non-Linux platforms or missing libc just log a warning and continue.
- Called from `_install_worker_signal_handlers()` after `signal.signal` calls, so every `worker_init` invocation picks it up automatically (pebble pool children only; the head process and `run_worker_manager` don't call `worker_init`).

Linux-only; this codebase runs on GKE and linux dev boxes so the fallback path is only hit in exotic test environments.

---


supersergiy (approved):
Good stuff

